PR #3, opened 2015-12-07
cleanup and cloud ready

Netherdrake:
I just deployed this into 'the cloud', using Docker.

You would probably want to remove .idea folder. Further refactoring and cleanups required.

Also, I see you have multiprocessing branch. I think it would be better to do asyncio, if you upgrade the project to Python 3.5, you can get async/await which is pretty sweet.


Netherdrake:
This is how it looks like:
https://gyazo.com/d93d1df54156c259783190d007cfe96e

utdrmac:
Would like to see this cleaned up. A dockerized deployment to AWS would be awesome.

PR #6, opened 2016-01-19
Fix Fload/Decimal compare in the config validation.

Webonaute:


Mikadily:
Thanks. Will test it tomorrow. Just interesting why ir didn't break for me before.

Webonaute:
it may be python version. I have Python 2.6.6.

I may have an unrelated question. does your script prevent lending dust? does it have any minimum amount config before create lending?

Mikadily:
Yes, if I understand your question correctly. Otherwise it would cause API error. It's been taken care of in: https://github.com/Mikadily/poloniexlendingbot/blob/master/lendingbot.py#L125

PR #8, opened 2016-01-23
Fixed a bug and added x-Days.

FrilledFish:
If Poloniex return a error the bot shuts down, i have added a 60 sec delay before checking again.
The bot will run longer before hitting a error now.


Mikadily:
X-days is a good idea.

But for the error, not sure how can it happen. Will have to look at it closer. My bot literally runs over a month without restart, without a problem. Perhaps you use Windows?

FrilledFish:
I use a raspberry pi, and it randomly gets to a error while running. I'm not 100% sure about it. It's says that my keys are invalid.

FrilledFish:
Don't merge yet, my branch have some bugs.

Nixtren:
Lately my bot has been catching Poloniex errors frequently. It seems their API sometimes gets overloaded. Glad to see someone implemented a check so I don't have to :P

Mikadily:
Please check another branch which is more robust: https://github.com/Mikadily/poloniexlendingbot/tree/multithreading

rnevet:
A timeout on requests has been added as part of this pull requests:
https://github.com/Mikadily/poloniexlendingbot/pull/69

I think this pull request needs to be rebased on the current master - with only the xdays change included and extended to include also command line parameter.

chfast:
I like the x-day feature.

FrilledFish:
I have made a new pull request. #108 
Look at that one instead of this.

PR #12, opened 2016-02-07
Implemented mass AutoRenew feature, set when bot stops and clear when…

rnevet:
Added ability to automatically switch Active Loans AutoRenew flag when it shuts down and clear the AutoRenew flag when it starts.
Useful if you can't keep the bot running for some reason.
The --clearAutoRenew command line parameter is useful if you want to stop loaning completely.

Both abilities are disabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,9 @@ gaptop = 200
 #If set to 0 all offers will be placed for a 2 day period
 sixtydaythreshold = 0.2
 
+# AutoRenew - if set to 1 the bot will set the AutoRenew flag for the loans when you stop it (Ctrl+C) and clear the AutoRenew flag when on started
+autorenew = 0
+
 #custom config per coin, useful when closing positions etc.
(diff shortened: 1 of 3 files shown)


Mikadily:
Nice feature. But not sure if time.sleep(sleepTime) move is necessary. Now if it ends in error it will rapidly retry.

Mikadily:
Oh I understand :) it won't catch a keyboardinterupt exception.

Mikadily:
It could be improved. But as it's disabled by default, won't hurt much. Thanks!

rnevet:
Yeah, I rushed a bit to implement it... in 2nd thought it would be better to just have 2 separate commands (either py files or cl params) for setting/clearing autorenew for all loans. I could make the change if it is useful.

rnevet:
Hmm.. and you are right, a sleep is needed also the exception case.

PR #13, opened 2016-02-10
Fixed nonce generation

firstrow:


Mikadily:
Can you briefly explain what problem it's addressing?

firstrow:
When I run original code I received error message:
"Nonce must be greater than 635715957374625288. You provided 1455121289279."

This code fixed the problem.

bhaak:
That error message means you have used the same API key with another program that used a different way of generating the nonce. Quote the documentation from Poloniex: "The nonce parameter is an integer which must always be greater than the previous nonce used".

Just generate a separate API key for the bot and you should be fine.

firstrow:
Thanks.

PR #23, opened 2016-03-03
Added "Average Lending Rate" calculation to status line.

rnevet:
Added a calculation of the average lending rate for all active loans to the status line.
I can't test it with multiple currencies as I'm not lending, but I don't see a reason why it shouldn't work.


Mikadily:
Appreciate your work, sir. Well done.

rnevet:
Thank you for the great script, been running on my phone now for weeks. :+1:

PR #32, opened 2016-03-22
Automatically adjust width of status line.

tom111:
The status line should be only one line.  Previously it had a fixed
width of 99 characters.  Now the width is set to the actual terminal
width.  This value is determined again every time the status line is
updated, so that a running lending bot can be accessed from terminals
with different widths, like in a tmux setup.

### DIFF
--- a/ConsoleUtils.py
+++ b/ConsoleUtils.py
@@ -1,0 +1,82 @@
+import os
+import shlex
+import struct
+import platform
+import subprocess
+
+def get_terminal_size():
+    """ getTerminalSize()
+     - get width and height of console
+     - works on linux,os x,windows,cygwin(windows)
+     originally retrieved from:
+     http://stackoverflow.com/questions/566746/how-to-get-console-window-width-in-python
+    """
+    current_os = platform.system()
+    tuple_xy = None
+    if current_os == 'Windows':
+        tuple_xy = _get_terminal_size_windows()
+        if tuple_xy is None:
+            tuple_xy = _get_terminal_size_tput()
+            # needed for window's python in cygwin's xterm!
+    if current_os in ['Linux', 'Darwin'] or current_os.startswith('CYGWIN'):
+        tuple_xy = _get_terminal_size_linux()
+    if tuple_xy is None:
+        print "default"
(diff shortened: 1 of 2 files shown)


tom111:
P.S. There was an unused local variable `now` in the function that I also removed.

Mikadily:
Well done. But I'm wondering how it would work on Windows?

Mikadily:
Hrrr Windows. I just found: http://stackoverflow.com/questions/566746/how-to-get-console-window-width-in-python
Will check on it later.

tom111:
I guess the options are:
- Require to python3.3 or higher and use `shutil.get_terminal_size`
- Import a bunch of code from the stackoverflow thread that works around it all
- On Windows, just put it to a fixed size like before.
- Just ignore Windows and merge.

Fornax96:
Windows's command line window is limited to 80 characters width right? Just hardcode it to 80 characters and no one will notice.

rnevet:
Windows cmd window is adjustable.

tom111:
This gist https://gist.github.com/jtriley/1108174 contains an implementation of that stackoverflow thread. Probably the only way to do it for all platforsm is to add  ~90 lines of code which branches for the different platforms.

Evanito:
I would not be against this extra code if it is necessary.

rnevet:
The original commit isn't compatible with windows.. before it can be merged we should at least support all platforms as now.

tom111:
OK, here is a new version based on the current HEAD and using the gist mentioned above. I have not tested this on Windows, please somebody try it. It seems to work well with Linux/OSX/tmux

rnevet:
Tested it on Windows 10, it seems to work fine with a slight modification - the number of characters that needs to be subtracted in https://github.com/tom111/poloniexlendingbot/blob/07cc7cb954e69970da14774060af3499c4d5c346/Logger.py#L24 is 4 and not 3 as you also need place for the

tom111:
OK, fine. Let's do -4 for everybody.

Mikadily:
Thanks @tom111 , good work.

PR #35, opened 2016-03-25
JSON Logger & web status page

rnevet:
Created a new logging option,  log to json file:
- status can be completely seen (including all coins)
- log size can be controlled
- can be parsed like in the lendingbot.html example

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -36,6 +36,14 @@ autorenew = 0
 #syntax: [COIN:mindailyrate:maxactiveamount, ... COIN:mindailyrate:maxactiveamount]
 #if maxactive amount is 0 - stop lending this coin. in the future you'll be able to limit amount to be lent.
 #coinconfig = ["BTC:0.18:1","CLAM:0.6:1"]
+
+#this option creates a json log file which includes the most recent status
+#uncomment both jsonfile and jsonlogsize to enable
+#jsonfile = www\botlog.json
+#limits the amount of log lines to save
+#jsonlogsize = 200
+#enables a webserver for the www folder, in order to easily use the lendingbot.html with the json log
+#startWebServer = true
 """
 
 loadedFiles = config.read([config_location])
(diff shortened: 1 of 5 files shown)


Mikadily:
Good job as always.
But I'm thinking how could we improve. It's a bit complicated to start using. How difficult is to start a webserver automatically on python to serve that html? So that it would be easy to use on headless server.

Mikadily:
And sorry for slow response.

rnevet:
I fixed a bug, it should load the botlog.json file automatically if it is in the same folder as the lendingbot.html and the browser allows it. (Internet explorer gives a warning that can be ignored, Chrome doesn't allow auto loading local files)

It is possible to start a local server using the following command (from the www folder):
'''
python -m SimpleHTTPServer 8000
'''
then the page will be available under:
http://127.0.0.1:8000/lendingbot.html

This is really dependent on the personal setup, i.e. if you are anyway running a webserver - you might not want to run another one.

rnevet:
I'll add an option to start a web server in the www folder.

Mikadily:
Awesome.

Now, I'm not so experienced with git. Maybe you know, how should I merge this update as a single commit? Or shouldn't I? It looks a bit ugly at the moment :)

rnevet:
Added WebServer option and squashed the commits locally, forced pushed to my repo - make it all nice looking.

lendingbot.py
```diff
@@ -231,32 +248,75 @@ def setAutoRenew(auto):
...
+			# quiet server logs
+			def log_message(self, format, *args):
+				return
+			# serve from www folder under current working dir
+			def translate_path(self, path):
+				return SimpleHTTPServer.SimpleHTTPRequestHandler.translate_path(self, '/www' + path)
+		
+		global server
+		server = SocketServer.TCPServer((HOST, PORT), QuietHandler)
+		print 'Started WebServer, lendingbot status available at http://'+ HOST +'/' + str(PORT) + '/lendingbot.html'
```
Mikadily: There's a mistake, ':' after HOST

Mikadily:
There's a mistake at the startup screen:

> Started WebServer, lendingbot status available at http://127.0.0.1/8000/lendingbot.html

And also the log file is not in the same dir as the html.

Otherwise it looks great, so impatient to merge it :)

Again sorry for me being slow.

rnevet:
Fixed the mistake in the startup message.
Make sure that the jsonfile config should be: www\botlog.json  - it wasn't in the first commit.

Mikadily:
Awesome. Happy with the result.

Issue #36, opened 2016-04-01
ERROR: 'provided'


Kryometric:
I don't know what happened, it says Welcome to Poloniex Lending Bot then the next line it says ERROR: 'provided'

rnevet:
Might be an api key issue, try replacing it with a new one.

martinkrung:
had the same problem, because '

apikey = 'XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX'
secret = 'd41d8cd98f00b204e9800998ecf8427ed41d8cd98f00b204e9800998ecf8427e'

is not working..

without quotes it works:

apikey = XXXXXXX-XXXXXXX-XXXXXXX-XXXXXXX
secret = d41d8cd98f00b204e9800998ecf8427ed41d8cd98f00b204e9800998ecf8427e

martinkrung:
Me again

I get this error in another case, because of the nonce:

            req['nonce'] = int(time.time()*1000)
change it to this and it works:

            req['nonce'] = int(time.time()*4222)

martinkrung:
here is a fix: https://github.com/Mikadily/poloniexlendingbot/pull/38

or use my repro: https://github.com/martinkrung/poloniexlendingbot

Webonaute:
It not fix the issue for me!

RDash21:
@Webonaute mind telling me which binary you are running?

FYI: You need the lendingbot.py and not poloniex.py :)

Webonaute:
Its my launch script I run in cronjob every minute : 

`#!/bin/bash
if [[ ! `pgrep -f "python lendingbot.py"` ]]; then
        echo "Application not found. Starting..."
        cd /home/poloniex/poloniexlendingbot; python lendingbot.py > poloniex.log 2>&1 &
        echo "done"
fi
`

Webonaute:
And the script was running fine in the last 2 month! 2 day ago, it stop working

RDash21:
@Webonaute try to remake the configuration file.

m33Lucky:
Fix #38 no working for me , apply for new api , with unrestricted access , works , 
If change to fix ip , error = provided .
Bot working now , but won't spread the lend , dumps max amount after mindailyrate

Webonaute:
I rewroted the python script in PHP finaly lol! now I can play a little more with the code!

Mikadily:
[...]

PR #38, opened 2016-04-05
higher nonce faktor to prevent "provided" error

martinkrung:
should fix this:

https://github.com/Mikadily/poloniexlendingbot/issues/36

for Python 2.7.9 on ubuntu

### DIFF
--- a/poloniex.py
+++ b/poloniex.py
@@ -41,7 +41,7 @@ class Poloniex:
             return json.loads(ret.read())
         else:
             req['command'] = command
-            req['nonce'] = int(time.time()*4222)
+            req['nonce'] = int(time.time()*4322)
             post_data = urllib.urlencode(req)
 
             sign = hmac.new(self.Secret, post_data, hashlib.sha512).hexdigest()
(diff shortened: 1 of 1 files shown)


Mikadily:
I'm not sure about it. Someone proposed it before and we agreed not to change nonce.
Here: https://github.com/Mikadily/poloniexlendingbot/pull/13

> Quote the documentation from Poloniex: "The nonce parameter is an integer which must always be greater than the previous nonce used".

Also pinging @bhaak

martinkrung:
I am not sure if it  has to do with the api key, as in #13. My running lendingbot was stopping and after setting a higher nonce it worked again. this makes no sense, as my clock is not running backwards. but I have to admit, that I do sync my different computer and If I start the lendingbot on both and not having the exact same time set this can result in this error

Webonaute:
I confiirm it not fixed the issue for me. I still have the error since 2 days

Evanito:
I find that this issue (also issue #36 ) was caused for me by having multiple applications using the same API key. My recommendation is to delete all APIs and start over with one key per integration.

That fixed it for me.

PR #48, opened 2016-04-16
Dynamically increase limit of returnLoanOrders request

rnevet:
This will prevent maxDailyRate loans because bot doesn't retrieve the entire loan list

### DIFF
--- a/Logger.py
+++ b/Logger.py
@@ -2,71 +2,96 @@ import sys
 import time
 import datetime
 import atexit
+import json
+import io
+from RingBuffer import RingBuffer
 
 class ConsoleOutput(object):
-    def __init__(self):
-        self._status = ''
-	atexit.register(self._exit)
+	def __init__(self):
+		self._status = ''
+		atexit.register(self._exit)
 
-    def _exit(self):
-        self._status += '  ' # In case the shell added a ^C
-        self.status('')
+	def _exit(self):
+		self._status += '  ' # In case the shell added a ^C
+		self.status('')
 
-    def status(self, status):
-        update = '\r'
-        update += status
-        update += ' ' * (len(self._status) - len(status))
-        update += '\b' * (len(self._status) - len(status))
-        sys.stderr.write(update)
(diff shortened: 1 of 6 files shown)


Mikadily:
I'm trying to digest. Can you briefly explain how you decide to increase the limit?

rnevet:
Basically if you reach the end of the offers and you are under the limit, it means you have seen all offers. and you can send at maxDailyRate. Otherwise there are more offers that you can fetch -> increase limit and repeat.

Mikadily:
Ok. I forgot, what's the default limit, 200? What if the lendbook size is exactly equal default limit?

rnevet:
Than it will increase once, no harm done.

rnevet:
The increase process is permanent per coin and per bot run... after it increases the bot will use that limit for the specific coin until it is restarted. (The idea at least)

Mikadily:
Ah. OK, interesting idea and looks good. I'll test and merge it. Hope in two days.

rnevet:
I tested it by raising the minDailyRate to 0.1 so for example for BTC the bot raised the limit to ~800 offers before it decided to make a loan at 0.1.

Mikadily:
I'm trying to think how to merge this, but it's a bit messy. Somehow it's intervened with your other commit (config file fix). Maybe you know a good way to squash it?
screenshot from 2016-05-05 16-57-19

rnevet:
You can either  cherry pick the commit locally instead of merging...  
Or you can merge it locally and then squash if needed.

rnevet:
Should I merge this change?
I can do the cherry picking.

Mikadily:
Yeah, go for it @rnevet. I just didn't want all the "patch-1" crap in the commit message. Otherwise I'm happy with how it works.

rnevet:
merged: https://github.com/Mikadily/poloniexlendingbot/commit/a81cbbdd71ee61ce28d0fb4866c4491ed2059788

Issue #54, opened 2016-05-05
Really minor "issue" - returnOpenLoanOffers doesn't match the API


Shtutnik:
returnOpenLoanOffers doesn't have any arguments, this is not really an issue..

PR #55, opened 2016-05-05
Matching returnOpenLoanOffers to the API

Shtutnik:
returnOpenLoanOffers doesn't have any arguments.
Fixes Issue #54


Mikadily:
Good find :) thanks

PR #58, opened 2016-05-08
loop through all possible currencies

Kurairaito:
this branch allow you to display performance of all your currencies lended.
It will create a row for each currency found in json file. if for any reason you won't display a currency, you can remove it from var currencies = [ at line 42

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -154,7 +154,7 @@ def refreshTotalLended():
 def stringifyTotalLended():
 	result = 'Lended: '
 	for key in sorted(totalLended):
-		result += '[%.3f %s @ %.4f%%] ' % (Decimal(totalLended[key]), key, Decimal(rateLended[key]*100/totalLended[key]))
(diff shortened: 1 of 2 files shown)


Kurairaito:
example : 
example

rnevet:
I like the idea of displaying the actual predicted earnings (definitely an improvement) but:

1. I think  using regex to harvest the information back from the string status is a wrong approach, It would be better to log the raw data in the json log. (would need to extend the Logger)
2. Why did you hard code the coin letters into the HTML? should be possible to parse it without, split the currencies based on [ and then parse each for the coin name and values.

IMO improving the logging to include raw data would give more flexibility and room for improvements.

Kurairaito:
Well I did it because it was easier. I could call poloniex api to grab currencies list also but it would bring more code and delay in retrieving data.

Improving logger seems also hard to me (doesn't have much knowlege in python)

Evanito:
LGTM

rnevet:
The improvement I suggested can be done later.

PR #62, opened 2016-05-12
Add Inactivemode

Evanito:
This is a major change and will require changes to default.cfg


rnevet:
Hi,

Configurations changes need to be also reflected in the Readme file.

Why convert the tabs to spaces? basically the entire file is marked as change makes it impossible to review.

Can you explain the goal of the changes?

rnevet:
It would be nicer if you wouldn't break current configs just for the sake of renaming "sleeptime" to "sleeptimeactive" and make the "sleeptimeinactive"  an optional parameter.

Evanito:
4 spaces is what I use, but I will revert and change my settings.

PR #65, opened 2016-05-13
Argument parsing

Evanito:
With this we can skip out on using a cfg file at all.

User can run the entire program using arguments (helps with automated programs with multiple users) without the need to have a cfg for each user.

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -1,4 +1,4 @@
-import io, sys, time, datetime, urllib2, json
+import io, sys, time, datetime, urllib2, json, argparse 
 from poloniex import Poloniex
 from ConfigParser import SafeConfigParser
 from Logger import Logger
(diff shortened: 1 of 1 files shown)


rnevet:
I don't actually understand the idea behind, the bot config is quite complex and most users have a hard time understanding it with the config file. I don't see someone running the bot completely from cmd line, I wouldn't want to either.

I could understand making the argument to override the config file.

Evanito:
The purpose of this is to add functionality, not remove it.
With this, we can have a csv file in the form
```
username,apikey,secret,ratemin,ratemax....
username2,apikey,secret,ratemin,ratemax....
username3,apikey,secret,ratemin,ratemax....
etc
```
That will allow people who want to run multiple bots with only one script to do so with very little hassle and a bit of coding.

People who do not want to use arguments will not have to, the config file will remain fully functional.

rnevet:
I see, my mistake I thought you removed the creation  of the default config file.

rnevet:
Do you want to mention this option in the readme?

rnevet:
@Evanito No mention in the readme?

Evanito:
@rnevet I plan on adding a link in the readme to a wiki page outlining the many arguments we have so that the readme isn't spammed.

rnevet:
@Evanito good idea. Should also setup a FAQ page, I'm not so experienced with WIKI pages.

Issue #40, opened 2016-04-09
[Suggestion] Include unlended coins in real "Average Lending Rate"


dutu:
Is it possible to have the "Average Lending Rate" calculation updated to show the **real** average?
Currently the average is only taking into consideration the active loans, and the funds waiting in open loan offers are excluded from the equation.

The correct way to calculate the average lending rate would be considering the total amount of coins available for lending (including those in the open loan offers).
Example: if one loan is active for 1 BTC at 0.1%, and 9 BTC waiting in open loan offers, currently the bot shows the average is 0.1%, when in fact the **effective**/**real** average is 0.01%.

rnevet:
Basically if you read the status line it says  "Lended:"  then you see the amount of coins currently lent and in effect everything in that line refers to the active loans and not the "total amount of coins" - so the Average Lending Rate is also for the lent amount.
i.e [2 BTC @ 0.1%] means you lent 2 BTC at an Average of 0.1%.

In my opinion showing the average calculation you asked for as part of the status line will be  misleading. Additionally If rates are over the minLendingRate the difference between the two calculations should be negligible.

What I noticed that really distorts the calculation is that sometime not all active loans are returned, making the calculation of the total lent and the rate just wrong - I opened a ticket for this and waiting a response from poloniex.

dutu:
@rnevet, thank you for your answer.
I understand perfectly when you are saying that the average shown is for the active loans (lent amount). Yes, the value shown is correct in this context.

However, it does not give the user an indication on how effective the bot is. Showing the "effective" rate (by taking into calculation the entire amount available for lending) is what gives indication on performance and user could possibly look into optimizing parameters (such as minLendingRate).

The topic of the discussion is "feature update" - it is not currently a fault in the code - but just that by having the feature updated to shown the **effective** rate is making it useful. Never mind it was just a suggestion :).

BTW, I'm not sure of what you are saying that "not all active loans are returned" - that should basically mean loan is still active, I haven't seen odd situations.

rnevet:
ok, I get what you are saying, do you have a suggestion how should the log look like so it isn't confusing?

dutu:
Implement these 2 points:
1. Change the name of what is shown today to "Average rate of active loans" 
2. Calculate and show the "Average lending rate" as noted in previous posts above

rnevet:
1. It is only mentioned in the commit comment which can't be changed.
2. As explained without other changes to the log message would be confusing.

[...]

PR #66, opened 2016-05-14
Improved logging with raw values

rnevet:
Added the ability to log raw status values (per coin) into the json log. (no need to perform regex on status strings to get values)
Added display of available funds and effective interest rate in lendingbot.html.

Should also give a solution to [Suggestion] Include unlended coins in real "Average Lending Rate" https://github.com/Mikadily/poloniexlendingbot/issues/40


rnevet:
I updated the calculation of multiple days earning interest rate - compound interest, if I understand correctly multi days interest rate should be Math.pow(<daily_rate>, <days>) and not multiply by the days. As the interest is daily.

Evanito:
LGTM :+1:

rnevet:
Are you sure that Poloniex doesn't calculate compound interest? makes the term daily interest rate a bit misleading.

Evanito:
No I don't know how Poloniex calculates "interest" but I assume it is simple daily, only compounding once the lend ends.

rnevet:
I added the estimation of BTC earnings for Alt coins based on polo ticker.
Reverted to compound interest calculation.

rnevet:
Received a response from Poloniex, they don't calculate compound interest on multi day loans. 
Merged.

Issue #61, opened 2016-05-11
Bot is not working sometimes


Antekk27:
From time to time I have the following effect with the bot:

- I start the bot
- the bot does his job like expect: cancels orders and places new ones.
- the bot goes on running, but is not carrying out any action. No output, no error is showing up.
- Having a look at poloniex lending area, I can see the the BTC Lending Balances is growing up slowly, which means that the bot 
as a matter of fact is not placing any offers.

I tested the bot under windows 7 on a private computer and under debian unix on a virutal server. Both installations shows the same effect from time to time.

Below you can see my config. There are still some DASH and STR loans open. With coinconfig = ["BTC:0.03:1","DASH:0.0:0","STR:0.0:0"] the bot should at least log "maxactive amount for ' + activeCur + ' set to 0, won\'t lend.'" every minute, but it doesn't

Kind Regards from Germany
Andreas

---

```
[BOT]
#sleep between iterations, time in seconds
sleeptime = 60
#minimum daily lend rate in percent
mindailyrate = 0.055
#max rate. 2% is good choice because it's default at margin trader interface. 5% is max to be accepted by the exchange
maxdailyrate = 2
#The number of offers to split the available balance uniformly across the [gaptop, gapbottom] range.
spreadlend = 3
#The depth of lendbook (in percent of lendable balance) to move through before placing the first (gapbottom) and last (gaptop) offer.
#if gapbottom is set to 0, the first offer will be at the lowest possible rate. However some low value is recommended (say 10%) to skip dust offers
gapbottom = 15
gaptop = 100
#Daily lend rate threshold after which we offer lends for 60 days as opposed to 2. If set to 0 all offers will be placed for a 2 day period
sixtydaythreshold = 0.06
#AutoRenew - if set to 1 the bot will toggle the AutoRenew flag for the loans when you stop it (Ctrl+C) and clear the AutoRenew flag when started
autorenew = 0
#custom config per coin, useful when closing positions etc.
#syntax: [COIN:mindailyrate:maxactiveamount, ... COIN:mindailyrate:maxactiveamount]
#if maxactive amount is 0 - stop lending this coin. in the future you'll be able to limit amount to be lent.
coinconfig = ["BTC:0.03:1","DASH:0.0:0","STR:0.0:0"]
#this option creates a json log file instead of console output which includes the most recent status
#uncomment both jsonfile and jsonlogsize to enable
#jsonfile = botlog.json
#limits the amount of log lines to save
[...]

PR #67, opened 2016-05-14
try to fix bug with connection timeout

Kurairaito:
This is an attemp to fix https://github.com/Mikadily/poloniexlendingbot/issues/61

In my case, the bot run on a Synology NAS. When I used download manager (on NAS), (wich take 100% of my bandwith), my bot was sometimes stuck too.

With this fix, I didn't experienced the same issue, but I'm not sure if it correct it completely

### DIFF
--- a/poloniex.py
+++ b/poloniex.py
@@ -53,8 +53,13 @@ class Poloniex:
                 'Key': self.APIKey
             }
 
-            ret = urllib2.urlopen(urllib2.Request('https://poloniex.com/tradingApi', post_data, headers))
-            jsonRet = json.loads(ret.read())
+            try:
+                ret = urllib2.urlopen(urllib2.Request('https://poloniex.com/tradingApi', None, headers),post_data,timeout=30)
+                jsonRet = json.loads(ret.read())
+            except Exception as e:
+                log.log("ERROR: " + str(e))
+                jsonRet = ""
+
             return self.post_process(jsonRet)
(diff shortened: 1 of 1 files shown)


Evanito:
LGTM :+1:

rnevet:
I don't think you should have exception handling inside there.

Evanito:
With the exception handling, won't it log an error saying what went wrong and then continue as normal, instead of causing failure? 
By having the exception set `jsonRet = ""` then we don't have an error where jsonRet is undefined or just mimics the last message.

rnevet:
There is already error handling in the main bot loop, which also includes a sleep so the bot doesn't hammer the server.
Returning "" will, I think, anyway crash the bot but in another place because this isn't a valid response for the request.

Evanito:
Hmm, you are probably right.

Close merge if you are sure.

Kurairaito:
Well on worst case, the bot crash and an other script can relaunch it.
Actually it is stuck at that line and you must monitor output file to see
if it is working or not, then kill it, and relaunch it. Maybe we could just
terminate the process if entering in exception ?

rnevet:
I noticed that you only added the timeout to the private API calls, I created another pull request with a more general solution that adds the timeout on the socket level so all requests are covered.
Here:
https://github.com/Mikadily/poloniexlendingbot/pull/69

PR #78, opened 2016-06-11
fix number display

Kurairaito:
fix display for number less than 100 satoshis displayed in scientific notation

### DIFF
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -101,7 +101,7 @@ function updateRawValues(rawData){
 			// print coin BTC earnings
             if(!isNaN(highestBidBTC)) {
                 var cell = row.appendChild(document.createElement("td"));
-                    cell.innerHTML = "<b>"+ couple +"<br/>highest bid:<br/>"+ highestBidBTC +"<b>";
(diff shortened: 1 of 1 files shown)


Kurairaito:
example : 

fix

same for highest bid (doge example)

rnevet:
I think it would be better to improve roundFloat function.

rnevet:
@Kurairaito 
My suggestion:
https://github.com/Mikadily/poloniexlendingbot/pull/79

Evanito:
Works as expected, cleared to merge!

Kurairaito:
Your suggestion doesn't works on my computer (win 10 up to date, neither edge nor chrome)

Evanito:
@Kurairaito His suggestion works fine for me. Screencap?

PS: html files are independent of OS, and only slightly dependent on browser, so if it works for us it should work for you.

rnevet:
@Kurairaito 
I basically built on your solution of adding toFixed and used in printFloat for every float, should work. 

Can you make sure that you checked out the correct branch "html_improvements" and javascript is updated?

Kurairaito:
Here is an example from my phone
screenshot_20160614-131812

rnevet:
@Kurairaito Thanks!
Corrected the implementation to include highestBid.

rnevet:
Implemented in https://github.com/Mikadily/poloniexlendingbot/pull/79

PR #79, opened 2016-06-12
fix display for number less than 100 satoshis

rnevet:
Replaced roundFloat with printFloat, now rounds and returns String with fixed precision.
Converted tabs to spaces.


Evanito:
Works as expected, cleared to merge!

rnevet:
Minor improvement:
Hide estimated earnings rows for coins with no active loans (sadly)

rnevet:
@Kurairaito can you confirm that this fixed the issue you found?

Kurairaito:
Yes it works,
Also the improvement to hide coin with no active loan works fine too

rnevet:
Thanks!

PR #86, opened 2016-07-05
Change sanity checks to make more sense

### LINKED ISSUE #82: I am having to disable 'sanity checks' minDailyRate & maxDailyRate
But by doing so, everything works as expected (I have specified my min and max as a percent, rather than a decimal).

### LINKED ISSUE #85: Fix "sanity checks" using conflicting percentages
The alert says "must be .003-5%", yet checks if the value is between .00003 and .05 

- [ ] Change sanity check values to percents not decimals
- [ ] Change default values to reflect these changes

Evanito:
Fixes #85
Closes #82

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -181,16 +181,16 @@ if config_needed:
 		pass
 sleepTime = sleepTimeActive #Start with active mode
 #sanity checks
-if sleepTime < 1 or sleepTime > 3600 or sleepTimeInactive < 1 or sleepTimeInactive > 3600:
-	print "sleeptime values must be 1-3600"
+if sleepTime <= 1 or sleepTime >= 3600 or sleepTimeInactive <= 1 or sleepTimeInactive >= 3600:
+	print "sleeptime values must be 1-3600 seconds"
 	exit(1)
-if minDailyRate < 0.00003 or minDailyRate > 0.05: # 0.003% daily is 1% yearly
+if minDailyRate <= 0.003 or minDailyRate >= 5: # 0.003% daily is 1% yearly
 	print "mindaily rate is set too low or too high, must be 0.003-5%"
 	exit(1)
-if maxDailyRate < 0.00003 or maxDailyRate > 0.05:
+if maxDailyRate <= 0.003 or maxDailyRate >= 5:
 	print "maxdaily rate is set too low or too high, must be 0.003-5%"
 	exit(1)
-if spreadLend < 1 or spreadLend > 20:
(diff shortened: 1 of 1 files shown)


lendingbot.py
```diff
@@ -181,16 +181,16 @@
...
-if minDailyRate < 0.00003 or minDailyRate > 0.05: # 0.003% daily is 1% yearly
+if minDailyRate <= 0.003 or minDailyRate >= 5: # 0.003% daily is 1% yearly
 	print "mindaily rate is set too low or too high, must be 0.003-5%"
 	exit(1)
-if maxDailyRate < 0.00003 or maxDailyRate > 0.05:
+if maxDailyRate <= 0.003 or maxDailyRate >= 5:
 	print "maxdaily rate is set too low or too high, must be 0.003-5%"
 	exit(1)
-if spreadLend < 1 or spreadLend > 20:
+if spreadLend = 1 or spreadLend >= 20:
```
rnevet: spreadLend = 1 seems wrong.

lendingbot.py
```diff
@@ -181,16 +181,16 @@
 		pass
 sleepTime = sleepTimeActive #Start with active mode
 #sanity checks
-if sleepTime < 1 or sleepTime > 3600 or sleepTimeInactive < 1 or sleepTimeInactive > 3600:
-	print "sleeptime values must be 1-3600"
+if sleepTime <= 1 or sleepTime >= 3600 or sleepTimeInactive <= 1 or sleepTimeInactive >= 3600:
+	print "sleeptime values must be 1-3600 seconds"
 	exit(1)
-if minDailyRate < 0.00003 or minDailyRate > 0.05: # 0.003% daily is 1% yearly
+if minDailyRate <= 0.003 or minDailyRate >= 5: # 0.003% daily is 1% yearly
```
rnevet: IMHO the values here should be 0.00003 and 0.05 since when we read the values from the configuration file we divide by 100.  i.e. if you configure the minDailyRate in the config to 0.003 it will be 0.00003 at this point. 

See lines 103, 105, 164, 165

Evanito:
I'm sitting in class right now and I don't have much time today, I just wanted to make a starting commit for this before I addressed the default config. Thanks for adding the DO NOT MERGE tag, I forgot.

rnevet:
IMHO there isn't a real issue with the logic (maybe some < where there should be a <=) but otherwise I think just the printed error is wrong... the config values aren't 0.03-5**%** but just 0.03-5 (without the percent).

PR #88, opened 2016-07-07
Update data from file every 60s

Evanito:
Without needing to refresh the page

### DIFF
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -185,9 +185,12 @@ function Timespan(name, multiplier) {
     };
 }
 
-$(document).ready(function () {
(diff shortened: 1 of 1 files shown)


Evanito:
No bugs when I was testing this.
What do you think @rnevet ?

rnevet:
I'm not so sure what exactly you mean, the data is already refreshed (every 30sec) from file without page refresh, see here:
https://github.com/Mikadily/poloniexlendingbot/blob/master/www/lendingbot.js#L158

Evanito:
I never noticed that when I was using it before, must be an id10t error.

Cheers

Issue #92, opened 2016-07-23
Bug: Bot cancels manually placed orders even if currency is ignored


yura-pakhuchiy:
I thought that I can use `coinconfig` for this, but bot cancel all manually placed orders even if `maxactiveamount` is set to 0.

rnevet:
@Evanito  was the maxactiveamount  thought as a way to disable a specific currency?

Evanito:
@rnevet Straight from the .cfg, "#if maxactive amount is 0 - stop lending this coin."

yura-pakhuchiy:
I think behaviour of `maxactiveamount`, is not a bug, it makes perfect sense if somebody want canel all loans for some coin just with bot (without using poloniex website). What I propose is to add new option to ignore some currency by bot, to let user manage that current himself (or use completelt different config/bot for that currency).

rnevet:
Actually it is a bug, the bot is on every cycle canceling all loan offers regardless of this flag. 
There is no reason why the bot would constantly cancel offers it didn't make.

It's also a relatively easy fix:
 https://github.com/Mikadily/poloniexlendingbot/blob/master/lendingbot.py#L286
Here where the bot loops and cancels all orders we should add a check if maxactiveamount = 0, and than not cancel the orders.

yura-pakhuchiy:
Such change will introduce inconsistent behaviour. Eg, user has 10 BTC.
1) If user sets `maxactiveamount = 0`, then bot will not cancel any orders placed by user and by bot previously.
2) If user sets `maxactiveamount = 2`, and lends 8 BTC manually. Bot will cancel all orders, placed by user and by bot, but will re-lend only 2BTC.

To make it consistent, for non-zero `maxactiveamount` values bot should remember which orders he placed and cancel only them.  I guess this is not a trivial change.

On the other hand, keeping present behaviour of `maxactiveamount` keep things consistent and introduction of new option fixes the problem I reported.

yura-pakhuchiy:
I edit my previous comment to be more clear. Just in case if you reading comments from email (email notifications only sent for the first version).

rnevet:
Actually from checking the code, `maxactiveamount` only disables making offers, it doesn't let you do partial lending (yet).
maxactiveamount  = 0  -> disabled.

This simplifies the issue to point 1, and also the solution.

yura-pakhuchiy:
Then documentation should be updated to state that non-zero values does not really affect lending ATM.

[...]

PR #93, opened 2016-07-25
BugFix: Bot cancels manually placed orders even if currency is ignored

rnevet:
Fix for issue #92


Evanito:
Nice and clean, looks good to me :+1:

PR #94, opened 2016-07-31
Make layout more mobile friendly

yura-pakhuchiy:
Current layout can not be displayed on mobile phone portrait view
without scrolling (320x... and 360x... resolutions) due to 5 rows. This
commit reduces maximum number of rows from 5 to 3 by combining
Active/Total coins and Average/Effective rate columns.

Other changes:
* Add yearly interest rate (with reinvestment and without). Can be used
  to compare interest rate with bank offers.
* Display percentage of lent coins from total coins.
* Replace `container-fluid` with `container` to improve readability on
  desktop screens.

### DIFF
--- a/www/lendingbot.html
+++ b/www/lendingbot.html
@@ -22,6 +22,10 @@
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js"></script>
     
     <script src="lendingbot.js" ></script>
+	
+	<style>
+		.inlinediv {display:inline-block;padding-right:5%;vertical-align:top}
+	</style>
 
 </head>
 
--- a/www/lendingbot.html
+++ b/www/lendingbot.html
@@ -28,7 +28,7 @@
 <body>
 
     <!-- Page Content -->
-    <div class="container-fluid">
+    <div class="container">
       <div class="row">
         <div class="panel-heading" style="background-color:#f2f2f2;display:flex;padding:0px" >
           <div class="col-xs-6">
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -71,14 +71,16 @@ function updateRawValues(rawData){
             var yearlyRate = effectiveRate * 365; // no reinvestment
             var yearlyRateReinv = (Math.pow(effectiveRate / 100 + 1, 365) - 1) * 100; // with daily reinvestment
             var lentPerc = lentSum / totalCoins * 100;
-            var poloRateText = '&nbsp;(<span title="Rate as seen on poloniex, before 15% fee.">poloniex</span>)';
-            var effRateText =  '&nbsp;(<span title="Effective rate, after poloniex 15% fee.">effective</span>)';
+            var avgRateText = '&nbsp;<span style="white-space:nowrap;" title="Average loan rate, simple average calculation of active loans rates.">Avg. (i)</span>';
+            var effRateText =  '&nbsp;<span style="white-space:nowrap;" title="Effective loan rate, considering lent precentage and poloniex 15% fee.">Eff. (i)</span>';
+            var compoundRateText =  '&nbsp;<span style="white-space:nowrap;" title="Compound yearly rate, the result of reinvesting the interest.">Comp. (i)</span>';
 
             var rowValues = [currency,
-                printFloat(lentSum, 4) + ' of ' + printFloat(totalCoins, 4) + ' (' + printFloat(lentPerc, 2) + '%)',
-                printFloat(averageLendingRate, 5) + '% / Day' + poloRateText + '<br/>' + printFloat(effectiveRate, 5) +
-                    '% / Day' + effRateText + '<br/>' + printFloat(yearlyRate, 2) + '% / Year (not reinvest)<br/>' +
-                    printFloat(yearlyRateReinv, 2) + '% / Year (reinvest)' ];
+                'Lent ' + printFloat(lentSum, 4) +' of ' + printFloat(totalCoins, 4) + ' (' + printFloat(lentPerc, 2) + '%)',
+                "<div class='inlinediv' >" + printFloat(averageLendingRate, 5) + '% Day' + avgRateText + '<br/>'
+					+ printFloat(effectiveRate, 5) + '% Day' + effRateText + '<br/></div>' 
+					+ "<div class='inlinediv' >" + printFloat(yearlyRate, 2) + '% Year<br/>'
+					+  printFloat(yearlyRateReinv, 2) + '% Year' + compoundRateText + "</div>" ];
 
             // print coin status
             var row = table.insertRow();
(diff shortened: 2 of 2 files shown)


rnevet:
With the new layout I need to scroll more vertically on PC and on my LG G4.
The old layout only meant a small scroll to the right (which usually wasn't needed as there are mostly titles there).

We should be able to make the layout more responsive using the bootstrap library... no reason not to use the horizontal screen space on desktop.

rnevet:
btw, I think the because of the unpredictability of loan rates, loan periods, loaned amounts etc.. displaying two values for yearly rates is useless.

yura-pakhuchiy:
Yes, I agree that layout should be totally redesigned to be more responsive. I tried to quickly fix mobile view without damaging desktop view much. New layout looks OK for me on desktop with 2 currencies, maybe with more currencies it is worse. New layout is definitely better for mobile, having horizontal scrolling is total disaster on mobile.

However I disagree that all horizontal space should be used on desktop (`containter-fluid`), that results in a lot of eye travel time, `container` is way easier to read.

Regarding yearly rates, I added them because for me (and I guess not only for me) is way easier to think that rate changes in range 4-10% yearly, than 0.0xxx% - 0.0yyy% daily.

yura-pakhuchiy:
However I understand that some people may prefer old layout. Maybe we can have both? And let users choose.

rnevet:
We can off course have multiple html layouts, however this would increase development effort for changes - i.e. you needed to add the yearly rate twice.

fluid vs non-fluid isn't the issue really... I have ~6 coins lent... since earnings take double the height now it takes 2-3 pages to see what's going on.

I agree about the convenience of yearly rate, just not sure we need to have 2 calculations.

yura-pakhuchiy:
All changes (except fluid) are in JS file and patch is rather small. We can pass argument to JS with desired layout and then apply my changes only for new layout.

yura-pakhuchiy:
All calculations and most of text will stay out of conditionals.

yura-pakhuchiy:
Two yearly rates also for convenience, maybe somebody regularly taking out money, so they need rate without reinvestment. Others would prefer reinvestment. For bigger daily rates they may differ significantly.

rnevet:
I think there is a miscalculation in the yearly interest.
image

yura-pakhuchiy:
Yearly interest is calculated based on effective daily interest, but in the
last picture something is definitely wrong with effective interest
calculation, it is much less than it should be.

rnevet:
Actually it isn't wrong,
First interest is the "Average loan interest" - which means the average of the loans.
The effective calculation takes in consideration the for this coin only 5% of the coins that I have are actually loaned.

yura-pakhuchiy:
I read your edited post, now I see why it is much lower

rnevet:
Sorry about that. :)

rnevet:
@yura-pakhuchiy I played a bit around with it on this branch:
https://github.com/Mikadily/poloniexlendingbot/tree/mobile

yura-pakhuchiy:
@rnevet No problem, and thank you for pointing out to me how effective amount is calculated. I have not really looked how it is calculated, just guessed from the name, my mistake.

Your changes look great! Keeps only vertical scrolling on mobile, but greatly improves desktop space efficiency. The only thing I think we should change is effective percentage calculation, it should not be affected by lent amount. It is very confusing, especially for yearly rate, it is not what users expect when they see interest rate. And of course we should still use lent amount for estimated earnings, I just propose to change effective rate calculation.

BTW, please do not reset author of original commit (it is marked as yours in your branch).

rnevet:
@yura-pakhuchiy 

Regarding effective rate:
https://github.com/Mikadily/poloniexlendingbot/issues/40
This is how effective rate was introduced. I would argue that the current calculation it is true to the word "effective" than disregarding the fact that coins in your lending account are actually not lent.

IMHO, The yearly calculation is misleading in any case as I mentioned "because of the unpredictability of loan rates, loan periods, loaned amounts etc.." it's a very very rough estimate.

Regarding the commit issue, I didn't so anything (just pulled your branch into a new one on this repo). I suggest you apply the changes back to your branch to have the commit on your name and part of the pull request.

yura-pakhuchiy:
@rnevet Done. Lets merge mobile layout since it is the main issue. I will write my thought on effective/yearly rate in separate PR/issue.

rnevet:
@Evanito Please take a look on the new layout before I merge it.

yura-pakhuchiy:
@rnevet I pushed small update to early rate calculation. It should not be affected by amount of lent coins ATM. I do not change "effective" rate calculation in any way, just change early rate. Hope you are OK with such change.

yura-pakhuchiy:
@rnevet Sorry for pushing after we agreed to discuss it in another issue. If you do not agree, just tell me and I will remove it from the branch to merge mobile layout as is.

rnevet:
I actually disagree, yearly estimation should include actual lent percentages - if you assume the momentarily average lending rate is good enough for estimating the yearly rate than the lent percentage is also good enough.

Basically I'm offering 10BTC for lending, but actually only 5BTC is lent - so my interest on the 10BTC will be only half of the interest for the 5BTC.  I might want to move my 10BTC somewhere else where I might manage to actually lend it in whole.

Additionally the lent percentage is effected by the minDailyRate, gaps and other config parameters - if for example I decrease the minDailyRate, I'll ensure that my coins are always lent - while average lending rate might decrease the yearly lending rate might increase.

yura-pakhuchiy:
@rnevet Just imagine you walk in the bank and say "I have 1000 USD and want to deposit some of them." and clerk says "OK, if you deposit all of them, then return rate would be 3%, but if you deposit 500USD return rate will be 1.5% (not because of different rate for different sums, but because you giving only half to us)". Nobody speaks like this, nobody expects that yearly return rate will be calculated like this. Rate is rate, and it does not depends on lent amount. It will be highly confusing otherwise.

rnevet:
@yura-pakhuchiy  Actually it's exactly like that in the bank, the interest rate depends on amount and period you are willing to give the bank. :) (thanks for the excellent point)

yura-pakhuchiy:
@rnevet yes, it depends on amount, but because of different they have different offerings for different amounts. But nobody multiples percent of deposit from your total savings on interest rate. It is completely does not make sense.

yura-pakhuchiy:
@rnevet I removed last commit from the branch as promised before, so we can merge as is. But please consider what I wrote above.

rnevet:
Let's move this issue to gitter chat: https://gitter.im/Mikadily/poloniexlendingbot

rnevet:
@yura-pakhuchiy, Imagine I have 100Euro in my Account, I can transfer it to my saving account and get 1% yearly for it. 
Now I have 1BTC in my account, how much percent will I get yearly for it if I transfer it to Lending? Well it depends if I actually manage to loan it (based on bot configuration).

So I need to take that in consideration if I want to compare the financial products - and that's what's important.

PR #100, opened 2016-08-10
Handle errors from Poloniex API.

denravonska:
Just a small fix which converts the errors returned from the Poloniex API to an exception. This allows you to get a better understanding of what went wrong, instead of getting a dictionary exception when the bot tries to access the data.

### DIFF
--- a/poloniex.py
+++ b/poloniex.py
@@ -5,6 +5,9 @@ import time
 import hmac,hashlib
 import socket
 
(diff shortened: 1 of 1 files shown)


rnevet:
Except you removed the return call in the returnTicker api call. :(

denravonska:
DOH, I tripple checked it. Sorry about that.

PR #102, opened 2016-08-11
Further optimization for mobile

rnevet:


rnevet:
@yura-pakhuchiy what do you think?

yura-pakhuchiy:
@rnevet I like all changes, except `Month` -> `M`. It affects all resolutions smaller than 768x..., but not really necessary even on 320x.. screen (it can be displayed without scrolling). However M/W/D/H is harder to understand than Month/Week/etc.

The only resolution it makes sense is 240x... (really small screens), but if you want to support them, then all other fields should be fixed (because they generate vertical scrolling now). If you will do it, I suggest to add separate media query for <320x... devices for this changes.

yura-pakhuchiy:
@rnevet In the first message, I missed that it also makes sense for non-BTC estimated earnings. I propose to apply new format only to non-BTC fields, but leave BTC fields with full form.

yura-pakhuchiy:
BTW, it is easier to review if commit message lists changes in the commit. Especially if there are several in one commit.

rnevet:
@Evanito @Mikadily

rnevet:
@yura-pakhuchiy Please take a look into the last update.
I removed the Month->M change - the resolution change is defined by the bootstrap framework, I don't want to define all the CSS myself.

yura-pakhuchiy:
@rnever I tested it already, looks good to me. Especially I like "add to home screen" feature and new icon. Great work!

yura-pakhuchiy:
@rnevet I tested it already, looks good to me. Especially I like "add to home screen" feature and new icon. Great work!

yura-pakhuchiy:
@rnevet BTW, what do you think about dropping table completely and using bootstrap's "col-xx-y" for formatting?

rnevet:
@yura-pakhuchiy it's a good idea (I also thought about it), just don't have time to do it. If you can go for it.

yura-pakhuchiy:
@rnevet minor glitch with new title bar: in local file mode it displays button over the status line

PR #108, opened 2016-09-11
The legend of xDays is back!

FrilledShark:
I have created a new version of my feature xDays. I have used this configuration for about a week. I copied the changes made to Github.

I'm not quite sure what I did. The script has evolved a bit since last time I used it. I just tried and tried until it was stable. I havn't observed any issues with it yet.
Feel free to look at it and see if it fits.

Old pull request: #8

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -139,7 +139,7 @@ if args.coinconfig:
 #End handling args.
 
 #Check if we need a config file at all (If all settings are passed by args, we won't)
-if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and args.xdaythreshold and args.xdays:
+if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and ((args.xdaythreshold and args.xdays) or args.sixtydaythreshold):
 	#If all that was true, we don't need a config file...
 	config_needed = False
 	print "Settings met from arguments."
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -83,6 +83,7 @@ parser.add_argument("-maxrate", "--maxdailyrate", help="Maximum rate you will le
 parser.add_argument("-spread", "--spreadlend", help="How many orders to split your lending into")
 parser.add_argument("-gapbot", "--gapbottom", help="Percentage of your order's volume into the ledger you start lending")
 parser.add_argument("-gaptop", "--gaptop", help="Percentage of your order's volume into the ledger you stop lending")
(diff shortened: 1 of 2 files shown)


Evanito:
LGTM! 👍

rnevet:
The change in logic is ok, however, documentation and backward compatibility are not taken care off.

FrilledShark:
I will try to fix backward compatibility. Is it ok if I just check if sixtydaythreshold is still present and if it is, use that instead of xdaythreshold?

rnevet:
Yeah that would be perfect, just don't want to break existing configs. If sixtydaythreshold  is present use it's value for the xdaythreshold with xdays=60?
Otherwise README also needs to be updated with the new config options.

FrilledShark:
I have tried to fix Backwards compatibility on my own branch. I will not put it on this pull request before I have tested it for bugs.
View my changes here: https://github.com/FrilledShark/poloniexlendingbot/commit/4847467f3fb3bcd483830fbdac19a92c7a3ef925

FrilledShark:
I have tried to make it backwards compatible. It works on my local copy, but try to see if I actually make it work.

FrilledShark:
Done.

lendingbot.py
```diff
@@ -132,7 +139,7 @@
 #End handling args.
 
 #Check if we need a config file at all (If all settings are passed by args, we won't)
-if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and args.sixtydaythreshold:
+if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and args.xdaythreshold and args.xdays:
```
rnevet: This means that sixtydaythreshold will not work as command line?

FrilledShark: I didn't know what to put there, so I just changed it from sixtydaythreshold to xdays...

rnevet:
I would say something like:
'''
((args.xdaythreshold and args.xdays) or args.sixtydaythreshold)
'''

lendingbot.py
rnevet: I would say something like:
```
((args.xdaythreshold and args.xdays) or args.sixtydaythreshold)
```

FrilledShark: Like this right? https://github.com/Mikadily/poloniexlendingbot/pull/108/commits/de313ae6541bf4187e3be312038ad5462f76031b

rnevet: I would say yes.

rnevet:
@Evanito, can you test this a bit before we merge it? I won't have access to my dev environment in the upcoming weeks.

FrilledShark:
It was running with my first pull request when I started it. I have updated changes now.

rnevet:
That's great, I would like to have 1 more person running it before merging. I would myself but can't at the moment.

PR #107, opened 2016-09-08
Catch exception when attempting to cancel loan offer

scrawl:
If a loan offer is taken some time in between the "returnOpenLoanOffers" and "cancelLoanOffer" calls, the attempt to cancel the order will raise an error since it's already taken.

When this happens, the uncaught exception propagates to the main loop and causes the bot to not lend out its funds for as long as sleepTime.

This issue will be more likely to occur the higher the "spreadLend" parameter is.

<pre>
2016-09-08 15:56:21 Canceling all BTC orders... Loan offer canceled.
2016-09-08 15:56:22 Canceling all BTC orders... Loan offer canceled.
2016-09-08 15:56:22 Canceling all BTC orders... Loan offer canceled.
2016-09-08 15:56:22 Canceling all BTC orders... Loan offer canceled.
2016-09-08 15:56:23 ERROR: Error canceling loan order, or you are not the person who placed it.
Traceback (most recent call last):
  File "lendingbot.py", line 461, in <module>
    cancelAndLoanAll()
  File "lendingbot.py", line 290, in cancelAndLoanAll
    msg = bot.cancelLoanOffer(cur,offer['id'])
  File "/home/scrawl/Dev/poloniexlendingbot/poloniex.py", line 157, in cancelLoanOffer
    return self.api_query('cancelLoanOffer',{"currency":currency,"orderNumber":orderNumber})
  File "/home/scrawl/Dev/poloniexlendingbot/poloniex.py", line 68, in api_query
    jsonRet = _read_response(ret)
  File "/home/scrawl/Dev/poloniexlendingbot/poloniex.py", line 38, in _read_response
    raise PoloniexApiError(data['error'])
PoloniexApiError: Error canceling loan order, or you are not the person who placed it.
</pre>

The fix is to catch exceptions when attempting to cancel orders, and assume the order is already filled if the cancellation fails.

<pre>
2016-09-08 16:01:38 Canceling all BTC orders... Loan offer canceled.
2016-09-08 16:01:38 Canceling all BTC orders... Loan offer canceled.
2016-09-08 16:01:39 Canceling all BTC orders... Loan offer canceled.
2016-09-08 16:01:39 Canceling all BTC orders... Loan offer canceled.
2016-09-08 16:01:40 Error canceling loan offer: Error canceling loan order, or you are not the person who placed it.
2016-09-08 16:01:41 Placing X BTC at Y% for 2 days... Loan order placed.
[...]

PR #112, opened 2016-09-29
Add the option to display the Account Estimated Earnings on BTC or in USD

prfiredragon:
I add the option to display the Account Estimated Earnings on BTC or in USD. The default is BTC.

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -405,8 +405,7 @@ def startWebServer():
 
 	try:
 		PORT = 8000
-		HOST = '0.0.0.0'
-		#'127.0.0.1'
+		HOST = '127.0.0.1'
 
 		class QuietHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
 			# quiet server logs
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -292,14 +292,8 @@ def cancelAndLoanAll():
 		for offer in loanOffers[cur]:
 			onOrderBalances[cur] = onOrderBalances.get(cur, 0) + Decimal(offer['amount'])
 			if dryRun == False:
-				try:
-					msg = bot.cancelLoanOffer(cur,offer['id'])
-					log.cancelOrders(cur, msg)
-				except Exception as e:
-					log.log("Error canceling loan offer: " + str(e))
-
-				#msg = bot.cancelLoanOffer(cur,offer['id'])
-				#log.cancelOrders(cur, msg)
+				msg = bot.cancelLoanOffer(cur,offer['id'])
+				log.cancelOrders(cur, msg)
 
 	lendingBalances = bot.returnAvailableAccountBalances("lending")['lending']
 	if dryRun == True: #just fake some numbers, if dryrun (testing)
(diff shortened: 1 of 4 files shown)


lendingbot.py
```diff
@@ -129,10 +133,14 @@
...
 		coincfg[cur[0]] = dict(minrate=(Decimal(cur[1]))/100, maxactive=Decimal(cur[2]))
+if args.nobtc:
+	onBTC = False
+else:
+	onBTC = True
 #End handling args.
 
 #Check if we need a config file at all (If all settings are passed by args, we won't)
-if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and args.sixtydaythreshold:
+if args.apikey and args.apisecret and args.sleeptimeactive and args.sleeptimeinactive and args.mindailyrate and args.maxdailyrate and args.spreadlend and args.gapbottom and args.gaptop and args.sixtydaythreshold and args.nobtc:
```
Evanito: Don't "require" an arg that has toggleable functionality.

lendingbot.py
```diff
@@ -283,8 +292,14 @@ def cancelAndLoanAll():
...
-				msg = bot.cancelLoanOffer(cur,offer['id'])
-				log.cancelOrders(cur, msg)
+				try:
+					msg = bot.cancelLoanOffer(cur,offer['id'])
+					log.cancelOrders(cur, msg)
+				except Exception as e:
+					log.log("Error canceling loan offer: " + str(e))
+
+				#msg = bot.cancelLoanOffer(cur,offer['id'])
+				#log.cancelOrders(cur, msg)
```
Evanito: Does this need to be changed?

Have you experienced errors with this code? It seems bundled in with the USD feature.

lendingbot.py
prfiredragon: done.

lendingbot.py
prfiredragon: I was testing (https://github.com/Mikadily/poloniexlendingbot/pull/107), I already remove it.

rnevet:
I would like to have a more generic approach, where we can offer to switch the currency in the html. Someone might be more interested in ETH or EUR (if available).

rnevet:
And I wouldn't name USDT -> USD, not exactly the same and no reason to.

Evanito:
That is a very good idea, you can change the config option to
```
outputCurrency = BTC #Options are BTC, USDT, ETH or anything as long as it has a direct BTC market.
```
Doing any other currencies that don't have a direct Poloniex market (Such as EUR) will require more overhead and we would need to agree on a source api, I recommend Preev.

rnevet:
Lets stick to Polo markets then.

prfiredragon:
Ok. Good idea, I am going began to test with BTC, USDT, ETH, XMR and DOGE.

rnevet:
Cool!

prfiredragon:
That was interesting. Done. The options are BTC, USDT and all in the BTC market.

Evanito:
I have reviewed the code and everything seems to be in order. Though with these kinds of changes I will want some user testing before merge. I will do it tomorrow if Raanan doesn't beat me to it. 
Great job and cheers!

prfiredragon:
Thank You. You are welcome.

PR #119, opened 2016-10-12
Create new cfg style that does not require generating.

Evanito:
## Description

This changes the config generation from within lendingbot.py to a separate file that only needs to be copied and renamed. 

Included is a fix for maxtolent, that changes 0.001 to the variable
minLoanSize

Cleans up readme.md to make account of the new docs and config file.

Changes donation address to multisig address. 

- Cleans up a huge waste of space in lendingbot.py
- Makes users who want to use an external host service not need python to generate the default config before they upload.

## TESTING STAGE

Not necessary, does not change bot's operation. Tested to ensure no bugs exist.

## Types of changes

## Checklist:

- [x] **My code adheres to the code style of this project**
- [x] **If my change requires a change to the config, I have updated the config file and command line arguments accordingly.**


Evanito:
@rnevet Really want your opinion on this.

rnevet:
Very nice, I'll try to review ASAP. 

It would be great for further development if we could some how not have duplicated configuration for command line and file, anny idea?

rnevet:
Ah and I would include the bug fix in a separate change, faster.

Evanito:
I meant to commit the bug fix right after merging maxtolent, but got carried away. Its a pretty foolproof change, only wasn't included because at the time I had forgotten we had a variable for minLoanSize.

I want to get this merged asap so we don't have a major change in structure when people are trying to contribute.

rnevet:
Should mention in the install instructions that Configuration is a step before running.

Evanito:
@rnevet Done

rnevet:
Actually, why not just copy the example file for the user if he ran the bot without a config file?

lendingbot.py
```diff
@@ -187,11 +117,8 @@
 	loadedFiles = config.read([config_location])
 	# Create default config file if not found
```
Evanito: Look at the total files changed instead of a single commit, I think that is why you are seeing the comment like that. (Because it currently says copy)

rnevet: My comment was before your last change

rnevet: My comment was before your last change

PR #127, opened 2016-10-31
Transform to full PEP8 conformity

Evanito:
## Description

Changes the entire bot to follow PEP8 formatting, which improves readability and enhances the maintainability of code. Does not change functionality at all.

Sorry about the huge diff, necessary evil for the long-term good this will do for us.

## TESTING STAGE

Bot still ran after these changes were made, that is a good sign to me. I will be testing certain configurations in the following days. I do not expect to test the arguments since Release 0.3 was the last version guaranteed to support it before we remove them soon.

## Types of changes

## Checklist:

- [x] **My code adheres to the [code style of this project.] (https://poloniexlendingbot.readthedocs.io/en/latest/contributing.html)**
- [N/A] **If my change requires a change to the config, I have updated the config file and command line arguments accordingly.**


Evanito:
Hmm, Okay need to change Hound's settings first.

Evanito:
So this is what It looks like right now. What do you think @rnevet? We can do this all at once or pull-by-pull and gradually change. After looking at this I honestly want to do the latter.

rnevet:
Hound violations are annoying, 79 chars line length?

rnevet:
btw, if you add ?w=1 to the compare view url it will filter most of the whitespace changes - makes it easier in this case.

lendingbot.py
```diff
@@ -26,562 +32,616 @@
...
+                coincfg[cur[0]] = dict(minrate=(Decimal(cur[1])) / 100, maxactive=Decimal(cur[2]),
+                                       maxtolent=Decimal(cur[3]), maxpercenttolent=(Decimal(cur[4])) / 100,
+                                       maxtolentrate=(Decimal(cur[5])) / 100)
+        except Exception as e:
+            print "Coinconfig parsed incorrectly, please refer to the documentation. Error: " + e
+            pass
+
+sleepTime = sleepTimeActive  # Start with active mode
+# sanity checks
+if sleepTime:
```
rnevet: This if could be inlined with AND in the next line

lendingbot.py
```diff
@@ -26,562 +32,616 @@
...
+    if sleepTime < 1 or sleepTime > 3600 or sleepTimeInactive < 1 or sleepTimeInactive > 3600:
+        print "sleeptime values must be 1-3600"
+        exit(1)
+if minDailyRate:
```
rnevet: can be inlined within next line

lendingbot.py
```diff
@@ -26,562 +32,616 @@
...
+    if minDailyRate < 0.00003 or minDailyRate > 0.05:  # 0.003% daily is 1% yearly
+        print "mindaily rate is set too low or too high, must be 0.003-5%"
+        exit(1)
+if maxDailyRate:
```
rnevet: can be inlined within next line

lendingbot.py
```diff
@@ -26,562 +32,616 @@
...
+    if maxDailyRate < 0.00003 or maxDailyRate > 0.05:
+        print "maxdaily rate is set too low or too high, must be 0.003-5%"
+        exit(1)
+if spreadLend:
```
rnevet: can be inlined within next line

lendingbot.py
```diff
@@ -26,562 +32,616 @@
...
+        try:
+            # parsed
+            coinconfig = (json.loads(config.get("BOT", "coinconfig")))
+            for cur in coinconfig:
+                cur = cur.split(':')
+                coincfg[cur[0]] = dict(minrate=(Decimal(cur[1])) / 100, maxactive=Decimal(cur[2]),
+                                       maxtolent=Decimal(cur[3]), maxpercenttolent=(Decimal(cur[4])) / 100,
+                                       maxtolentrate=(Decimal(cur[5])) / 100)
+        except Exception as e:
+            print "Coinconfig parsed incorrectly, please refer to the documentation. Error: " + e
```
rnevet: Not related to this change, but shouldn't the bot stop if the config is bad?

rnevet:
We should also decide if variables are came cased or underscored, right now it's a mix. 

I say lets make it once, but let make it good.

lendingbot.py
Evanito: I did not change logic in this PR, it was just marked a violation that we had an exception handler that did nothing so I added a print to notify the user.

Evanito:
@rnevet PEP8 says that global variables are camelCase and function variables are lower_case_with_underscore. That way, we can never have a function variable that accidentally overwrites a global variable, and we will always know which type we are using when writing a new function.

lendingbot.py
Evanito: I'll be honest, I don't like the sanity checks. I feel like they need to be done during parsing itself or right before the respective variable is used (I recommend the former). Right now, PyCharm sees a logical possibility that the sanity checks could run without first defining the variables and throws several errors. I would like to get that changed asap.

lendingbot.py
rnevet: I know... if you agree we should open an issue for this.

rnevet:
Are you sure about global variables? 
https://www.python.org/dev/peps/pep-0008/#global-variable-names
"The conventions are about the same as those for functions."

Evanito:
No I am not sure, it was just what PyCharm enacted. That means I need to make more changes. Ill get at it later today.

rnevet:
I'll do it, I have a couple of minutes.

rnevet:
lgtm

Evanito:
👍

PR #136, opened 2016-11-09
Remove arguments

### LINKED ISSUE #128: Remove Most Arguments
As seen in Release 0.3, we will be discontinuing arguments very soon. This is due to them creating a hassle where anyone who wants to implement a feature or option basically has to do the config part of it twice, which is not fun.

We will be keeping --config and --dryrun because those are the only two options that cannot be changed inside a config. If you wish to run multiple instances , all you need to do is run it with several configs instead of from a huge unorganized terminal line.

This change will happen very soon.

Evanito:
## Description

Removed all arguments that can be configured in a cfg file, left ``--config`` and ``--dryrun``

Closes #128 

This is to make it easier to add new features in the future by not requiring contributors to program adding the configuration logic for the feature twice. 
## Types of changes

## Checklist:

- [x] **If my change requires a change to the config, I have updated the config file and command line arguments accordingly.**

Removed arguments but made sure the remaining logic was the same.


Evanito:
I went back and disabled checking variable name conventions until we get them worked out, so Codacy shouldn't do this again.

lendingbot.py
```diff
@@ -221,15 +121,10 @@ def timestamp():
 log = {}
 
 # check if json output is enabled
-jsonOutputEnabled = (config.has_option('BOT', 'jsonfile') and config.has_option('BOT', 'jsonlogsize')) or (
-    args.jsonfile and args.jsonlogsize)
+jsonOutputEnabled = (config.has_option('BOT', 'jsonfile') and config.has_option('BOT', 'jsonlogsize'))
```
rnevet: valid issue.

lendingbot.py
```diff
@@ -15,187 +15,87 @@
...
+if config.has_option("BOT", "coinconfig"):
+    try:
+        # parsed
+        coinconfig = (json.loads(config.get("BOT", "coinconfig")))
+        for cur in coinconfig:
+            cur = cur.split(':')
+            coincfg[cur[0]] = dict(minrate=(Decimal(cur[1])) / 100, maxactive=Decimal(cur[2]),
+                                   maxtolent=Decimal(cur[3]), maxpercenttolent=(Decimal(cur[4])) / 100,
+                                   maxtolentrate=(Decimal(cur[5])) / 100)
+    except Exception as e:
```
rnevet: we can consider reducing min variable length to 1 (or renaming Exceptions to ex)
To reduce min variable name to 1:
[a-z_][a-z0-9_]{2,30}$ -> [a-z_][a-z0-9_]{*1*,30}$

rnevet:
image
Basically this regex needs to be change to lower cased:
[a-z_][a-z0-9_]{2,30}$

Otherwise it lgtm, needs a testing period.

lendingbot.py
Evanito: I am not opposed to reducing minimum length to 1, I just followed Mik's exception style

Evanito:
LGTM

rnevet:
Lgtm

PR #143, opened 2016-11-20
Max to lend code complexity

rnevet:
My suggestion to reduce the code complexity of the amount_to_lend function.

Minor variable renaming in Logger


Evanito:
LGTM

PR #171, opened 2017-01-06
Hide coins if mindailyrate not met

### LINKED ISSUE #46: Feature Request - Leave coins in 'My Balance' if min lend rate not reached
Rather than have our coins siting in the lending ledger at the max rate, leave them in 'My Balance' section. This would deflate the ledger, and lower the server hit on poloniex adding and removing coins ever XX seconds. Deflating the lending ledger could entice margin traders to borrow at higher rates instead of waiting for everything to continue getting lower and lower.

At the moment I have coins being added and removed every 60 seconds at .19%

Instead the lendingbot could wait until a users minimum lending rate is reached and then the coins would be added to the lending ledger. A % threshold could be added to catch an uptrend.

10% threshold example at 0.1% min lend rate: 

when lending rate reaches 0.09% lending bot will start adding coins to be lent out at .1% and up.

Evanito:
## Description

If a coin's market does not meet its mindailyrate, then it will wait instead of lending at min.

Closes #46

## Types of changes

## Checklist:

### DIFF
--- a/default.cfg.example
+++ b/default.cfg.example
@@ -47,6 +47,9 @@ minloansize = 0.001
 #If this happens, KeepStuckOrders will keep your order where it is so maybe it can be filled. Otherwise it will be canceled and held until orders expire.
 keepstuckorders = True
 
+#Hide coins - Instead of keeping your coins lent out at minlendrate when it is not met, the bot will hold them and wait for the rate to surpass it.
+hideCoins = True
+
 #End date for lending, bot will try to make sure all your loans are done by this date so you can withdraw or do whatever you need.
 #Uncomment to enable.
 #Format: YEAR,MONTH,DAY
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -136,6 +136,14 @@ Very few situations require you to change these settings.
 - A "Stuck" order occurs when it partially fills and leaves the coins balance total (total = open orders + let in balance) below your ``minloansize`` and so the bot would not be able to lend it again if it was canceled.
 - When disabled, stuck orders will be canceled and held in balance until enough orders expire to allow it to lend again.
 
+``hideCoins`` If True, will not lend any of a coin if its market low is below the set ``mindailyrate``.
+
+- Default value: True
+- Allowed values: True or False. Commented defaults to True
+- This hides your coins from appearing in walls.
+- Allows you to catch a higher rate if it spikes past your ``mindailyrate``.
+- Not necessarily recommended if used with ``analyseCurrencies`` with an aggressive ``lendingStyle``, as the bot may miss short-lived rate spikes.
+
 ``endDate`` Bot will try to make sure all your loans are done by this date so you can withdraw or do whatever you need.
 
 - Default value: Disabled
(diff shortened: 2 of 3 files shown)


Evanito:
I considered having it not hide if it hits an artificial min set by the Analysis module, but decided against it.

rnevet:
It will work, but seems a bit aggressive - any offer below min rate will automatically disable the bot for that currency. 

Another note using `loans['offers'][0]` is risky, I already encountered an empty offers book - which led to a crash since there was no object in position 0.

default.cfg.example
```diff
@@ -47,6 +47,9 @@ minloansize = 0.001
 #If this happens, KeepStuckOrders will keep your order where it is so maybe it can be filled. Otherwise it will be canceled and held until orders expire.
 keepstuckorders = True
 
+#Hide coins - Instead of keeping your coins lent out at minlendrate when it is not met, the bot will hold them and wait for the rate to surpass it.
+hideCoins = True
```
rnevet: isn't this against the new naming conventions? should be all lower case?

docs/configuration.rst
```diff
@@ -136,6 +136,14 @@ Very few situations require you to change these settings.
 - A "Stuck" order occurs when it partially fills and leaves the coins balance total (total = open orders + let in balance) below your ``minloansize`` and so the bot would not be able to lend it again if it was canceled.
 - When disabled, stuck orders will be canceled and held in balance until enough orders expire to allow it to lend again.
 
+``hideCoins`` If True, will not lend any of a coin if its market low is below the set ``mindailyrate``.
+
+- Default value: True
+- Allowed values: True or False. Commented defaults to True
+- This hides your coins from appearing in walls.
+- Allows you to catch a higher rate if it spikes past your ``mindailyrate``.
+- Not necessarily recommended if used with ``analyseCurrencies`` with an aggressive ``lendingStyle``, as the bot may miss short-lived rate spikes.
```
rnevet: I think in general this when this flag is set to true you might miss spikes.

modules/Lending.py
```diff
@@ -171,6 +173,9 @@ def lend_cur(active_cur, total_lended, lending_balances):
 
     loans = api.return_loan_orders(active_cur, loanOrdersRequestLimit[active_cur])
     loans_length = len(loans['offers'])
+    if hide_coins and Decimal(loans['offers'][0]['rate']) < Decimal(cur_min_daily_rate):
```
rnevet: loans['offers'][0] might throw an exception if no offers are returned from server (i.e. offers list is empty, which I saw happen during a spike in loan demands on XMR)

docs/configuration.rst
Evanito: Yeah, that is just a situation where I saw it might cause the most trouble

default.cfg.example
Evanito: At some point we will rename all the configuration settings, since they are a 50/50 split between lowercase or camelCase. 

Since it will be a breaking change we will need to enforce conventions after that point, but right now there is no point since there is no consistency.

modules/Lending.py
Evanito: Will solve this in rewrite, if issues starting coming up soon I may put a try-except around it.

Evanito: Fixed in rewrite

PR #174, opened 2017-01-07
Overhaul lending module

### LINKED ISSUE #165: Not enough XXX available to offer.
I'm seeing several errors like this one, I'm not sure if this is because of recent logic changes. 
Running on develop branch, but I think they also happen on master.

Example:

2016-12-22 16:27:14 Error: Not enough FCT available to offer. Requesting createLoanOffer
2016-12-22 16:27:14 Placing 2.07942785 FCT at 0.019899% for 2 days... Loan order placed.
2016-12-22 16:27:13 Placing 1.03971393 FCT at 0.019799% for 2 days... Loan order placed.

### LINKED ISSUE #180: Required minloansize changed (Amount must be at least 0.01)
Hi Guys,

Poloniex has changed the minimum loan amount of BTC (0.01), Clam (1.0) and other coins. This is causing an error in the BOT.

Thank you

### LINKED ISSUE #188: API request rate limit.
Apply a rate limit for api requests that only allows a maximum of 6 requests to be sent per second, so as to not go over Poloniex's limit.

Related issue: #167

### LINKED ISSUE #192: Traceback Error 123
Hey! Thank you so much for building this project.  It has helped a ton with setting up my lending. I came across this issue when trying to run the program today. It has been working great up until this point. Not sure what has changed.

Traceback (most recent call last):
  File "lendingbot.py", line 18, in <module>
    os.chdir(os.path.dirname(sys.argv[0]))  # Allow relative paths
WindowsError: [Error 123] The filename, directory name, or volume label syntax is incorrect: ''

### LINKED ISSUE #195: Custom mindailyrate not working.
On `lending-rewrite` branch, bot is not respecting custom coin config:

> 2017-01-19 17:15:53 Placing 6.11105483 MAID at 0.34989% for 2 days... Loan order placed.
> 2017-01-19 17:15:52 Using custom mindailyrate 0.400% for MAID

Latest `lending-rewrite` (2c3e4c5130a913808c5deb3aeae59079beb86a1d)
Debian 8
Python 2.7.9

Evanito:
Goal: To overhaul the lending module that has become difficult to modify, will allow us to fine-tune its function in the future.

Made much of the creation of orders more direct.

Fixes #165 - This error was due to splitting a satoshi in half, and they both wanted to round up.
Fixes #180 - Poloniex minloansize was increased, bot now dynamically changes to any potential increases.
Fixes #188 - Governs max API requests per second
Fixes #192 - Chdir now checks for lendingbot.py instead of an accidentally deletable file
Fixes #195 - Bugfixes about coincfg settings and the like

## Types of changes

## Checklist:

### DIFF
--- a/modules/Data.py
+++ b/modules/Data.py
@@ -97,7 +97,7 @@ def update_conversion_rates(output_currency, json_output_enabled):
         if output_currency == 'BTC':
             log.updateOutputCurrency('highestBid', '1')
             log.updateOutputCurrency('currency', output_currency)
-        else:
+        elif output_currency != 'USDT':
             url = "https://blockchain.info/tobtc?currency={0}&value=1".format(output_currency)
             try:
                 highest_bid = json.loads(urlopen(url).read())
--- a/modules/Data.py
+++ b/modules/Data.py
@@ -1,6 +1,8 @@
 import datetime
 import time
 from decimal import Decimal
+from urllib import urlopen
+import json
 
 api = None
 log = None
@@ -95,6 +97,15 @@ def update_conversion_rates(output_currency, json_output_enabled):
         if output_currency == 'BTC':
             log.updateOutputCurrency('highestBid', '1')
             log.updateOutputCurrency('currency', output_currency)
+        else:
+            url = "https://blockchain.info/tobtc?currency={0}&value=1".format(output_currency)
+            try:
+                highest_bid = json.loads(urlopen(url).read())
+                log.updateOutputCurrency('highestBid', highest_bid)
+                log.updateOutputCurrency('currency', output_currency)
+            except ValueError:
+                print "Currency {0} is not valid, choose a valid currency from here: {1}"\
+                      .format(output_currency, "https://blockchain.info/api/exchange_rates_api")
 
 
 def get_lending_currencies():
--- a/modules/Data.py
+++ b/modules/Data.py
@@ -109,7 +109,8 @@ def get_lending_currencies():
 
 
 def truncate(f, n):
-    '''Truncates/pads a float f to n decimal places without rounding'''
+    """Truncates/pads a float f to n decimal places without rounding"""
+    # From https://stackoverflow.com/questions/783897/truncating-floats-in-python
     s = '{}'.format(f)
     if 'e' in s or 'E' in s:
         return '{0:.{1}f}'.format(f, n)
--- a/modules/Data.py
+++ b/modules/Data.py
@@ -106,3 +106,12 @@ def get_lending_currencies():
     for cur in lending_balances:
         currencies.append(cur)
     return currencies
+
+
+def truncate(f, n):
+    '''Truncates/pads a float f to n decimal places without rounding'''
+    s = '{}'.format(f)
+    if 'e' in s or 'E' in s:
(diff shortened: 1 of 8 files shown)


rnevet:
Completely hiding timeout errors is not recommended, they can indicate an issue that needs to be taken care of.

Evanito:
What do you recommend? I was thinking either a console message saying it had to retry, or possibly a limit of retries per one request before throwing? What if their internet is down? Our users are not as likely to understand the error is just due to poor connectivity and so should be ignored. Retrying it gracefully is much better than what we have now, which completely skips the request if it throws and thus ends up with problems like #181

rnevet:
I think a console message should be enough.

Evanito:
Due to the nature of the changes in this PR, I strongly suggest we merge this soon.
I can do a rewrite of the cancel logic in another PR.

rnevet:
I wouldn't hide any of these errors, log them as Retrying Due to: <original error message>. 
Also why not handle them in the external error handling? with existing sleep.

BTW, timeout is already handled here:
https://github.com/Mikadily/poloniexlendingbot/blob/master/lendingbot.py#L96

rnevet:
@Evanito  I'm not sure you fixed issue #181, or how do you figure you did?

Evanito:
The problem with handling retries in the main exception handle is that it doesn't actually retry. it just skips the rest of the run then starts over. a88e03d allows it to retry right where it left off.

rnevet:
I'm not sure if retrying timeouts, internal errors after some seconds of sleep without checking if something has changed is a good idea anyway. The idea was "something went wrong" restart.
In any case also internal error should be logged as a warning.

Evanito:
Planning on merging.
@rnevet review?

rnevet:
I'll try to review this ASAP.

modules/Data.py
```diff
@@ -106,3 +106,12 @@ def get_lending_currencies():
     for cur in lending_balances:
         currencies.append(cur)
     return currencies
+
+
+def truncate(f, n):
```
Evanito: It is from the ever helpful stackexchange https://stackoverflow.com/questions/783897/truncating-floats-in-python

Would you suggest I instead multiply the number by 100,000,000, floor(), then divide by 100,000,000.0?

rnevet: No, keep the function as is, just add the link in a comment so we know where it came from an what the logic is. Not a simple one to understand by just reading it.

Evanito: Okay I will add that as a comment, wait for approval of laxdog's PRs, then merge this.

Evanito:
Must fix #195 before merging.

modules/Lending.py
```diff
@@ -278,7 +278,7 @@ def lend_cur(active_cur, total_lended, lending_balances):
     orders = construct_orders(active_cur, active_bal)  # Construct all the potential orders
     i = 0
     while i < len(orders['amounts']):  # Iterate through prepped orders and create them if they work
-        below_min = Decimal(orders['amounts'][i]) < Decimal(cur_min_daily_rate)
+        below_min = Decimal(orders['rates'][i]) < Decimal(cur_min_daily_rate)
```
Evanito: D'oh!

Issue #197, opened 2017-01-20
Being able to show bitcoin value in mBTC or bits. 


tuberculo:
As bitcoin value is relatively high, the amount in bitcoin someone posses tends to be a low value. To me, using all those decimal places is very unnatural, so I prefer to see the amounts in bits (µBTC). 
I suggest an option to show the amount of bitcoin in mBTC or bits, as the user prefers.

Evanito:
Shouldn't be too complicated with #149's existing functionality.
@rnevet?

rnevet:
Should be easy enough to add as a Get parameter to the page.

rnevet:
@tuberculo you can test the implementation on https://github.com/Mikadily/poloniexlendingbot/tree/custom_btc_display_unit
branch

rnevet:
Implemented and merged.

tuberculo:
I tested and it is working great. Thanks.
But I think that the total lent should also be in the selected unit.

rnevet:
Good point.

PR #204, opened 2017-01-29
Output Currency revolution

rnevet:
Display estimated earnings for all lent coins in output currency
Display output currency exchange rate for all lent coins (inspired by @yura-pakhuchiy)
Display total lent for BTC respecting display units (mBTC, Bits, Satoshi) fix comment on https://github.com/Mikadily/poloniexlendingbot/issues/197

## TESTING STAGE
In progress

## Types of changes

## Checklist:


rnevet:
@yura-pakhuchiy  what do you think?

yura-pakhuchiy:
@rnevet I'm away from laptop right now. I will try this branch in 2 days and let you know

rnevet:
add the min limit rate to the log when bot decides not to lend. https://github.com/Mikadily/poloniexlendingbot/issues/212

yura-pakhuchiy:
@rnevet Looks good! Few notes:

1.  `1 BTC = 1012.0000 USDT`

Looks like too many zeroes after dot. I think there should be configuration option for output currency precision.

2. I'm not sure that I like all currencies converted to USDT. Would be great to have option which preserves old behavior (all crytptos converted to BTC and only summary converted to output currency)

rnevet:
Thanks, regarding your points,

`1 MAID = 0.1359 EUR` , `1 BTC = 941.0968 EUR` - the more configuration options we are adding, it makes the bot harder to maintain. I could reduce precision to 3 and we can trim zeros on right side. otherwise it doesn't seem to me worth the effort.

yura-pakhuchiy:
I agree with "the more configuration options we are adding, it makes the bot harder to maintain", but there is another side - features makes software useful.

In this particular case, there is no single precision which will suit all possible summary currencies. 2-4 is enough for USD/EUR, but if we display eg., STR in BTC, it is definitely not. 1 STR = 237 satoshi ATM.

This option does not add a lot to maintenance burden, just get one value and pass it to `printFloat` directly without any manipulations. I do not propose to add precision option per every coin user holds, just one option for summary coin.

rnevet:
I actually think it is solvable with logic, something in the direction of show a min of 2 precision and more if value is lower, basically always 2 numbers to the right of the most significant one. What do you think?

rnevet:
Do you have maybe time to implement the switch option that you suggested in 2. ?  Use BTC or Output for inlime coin summary?

rnevet:
@yura-pakhuchiy I think I solved the exchange rate display issue, check it out.

rnevet:
I actually like the prettyFloat(value, 2) display, I think we should use it everywhere...

Evanito:
When set to Bits, I get this exchange rate shown for USDT: `1 Bits = 1041.2000 USDT`

rnevet:
Latest version? As that also doesn't look like pretty Float.

Evanito:
```
pi@evpi:~/poloniexlendingbot $ sudo git branch
* custom_btc_display_unit
  master
pi@evpi:~/poloniexlendingbot $ sudo git pull
Already up-to-date.
```

rnevet:
I'll check. :)

PR #221, opened 2017-02-07
Make sure we always have an ouput currency if blockchain.info is down…

laxdog:
…, or an incorrect currency is used

Hit a problem earlier when I couldn't connect to blockchain. This was mentioned in the original pull request and I forgot to fix it. 

## Description
Use BTC as the output currency if user enters the wrong value, or we can't connect to blockchain.info

## TESTING STAGE
Testing now

## Types of changes

## Checklist:


modules/Data.py
```diff
@@ -114,6 +114,11 @@ def update_conversion_rates(output_currency, json_output_enabled):
                 log.log_error("Failed to find the exchange rate for outputCurrency {0}!".format(output_currency))
                 log.log_error("Make sure that {0} is either traded on Poloniex or supported by blockchain.info: {1}"\
                               .format(output_currency, "https://blockchain.info/api/exchange_rates_api"))
+                log.log_error("Using BTC as the output currency".format(url))
```
rnevet: .format(url) ?

modules/Data.py
```diff
@@ -114,6 +114,11 @@ def update_conversion_rates(output_currency, json_output_enabled):
...
+                update_conversion_rates('BTC', json_output_enabled)
```
rnevet: can just be replaced with:
```
             log.updateOutputCurrency('highestBid', '1')
             log.updateOutputCurrency('currency', 'BTC')
```

modules/Data.py
```diff
@@ -114,6 +114,11 @@ def update_conversion_rates(output_currency, json_output_enabled):
...
+            except:
```
rnevet: 1. why do we need an additional except handler?
2. see previous comments.

modules/Data.py
laxdog: Typo, forgot to remove it when I copied the log line. I'll fix that.

modules/Data.py
laxdog: 1. Because otherwise we get a log message saying to open a github issue to fix something we can't fix (Connection error usually). The first one catches ValueError which occurs when you put a wrong currency value in. The second catches everything else. 

2. As above, I can change it if you like.

modules/Data.py
laxdog: I actually thought calling the function again was cleaner. It means if it changes in the future we don't need to change that logic. 

Though I can change it if you prefer.

rnevet: I don't like calling the function again as it does a call to polo backend and updates all rates.  
However, I have an idea, we should set the BTC out by default by changing this:
```
         if output_currency == 'BTC':
             output_currency_found = True
             log.updateOutputCurrency('highestBid', '1')
             log.updateOutputCurrency('currency', output_currency)
```
to:
```
         log.updateOutputCurrency('highestBid', '1')
         log.updateOutputCurrency('currency', 'BTC')
         if output_currency == 'BTC':
             output_currency_found = True
```

laxdog: Yea, OK. That's a good fix. Making the changes now.

Issue #205, opened 2017-01-30
BadStatusLine:  Requesting returnLoanOrders


laxdog:
Could just have been a time out, I'm not sure. I've not used used the market analysis stuff much. Thought I'd raise it anyway, maybe others are seeing it.

Running from lending-rewrite @ a2f14e451866ea7b70f1cfa46fbc0201b87ecf12

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 811, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/threading.py", line 764, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/dog/poloniexlendingbot/modules/MarketAnalysis.py", line 57, in update_market_loop
    update_markets()
  File "/home/dog/poloniexlendingbot/modules/MarketAnalysis.py", line 66, in update_markets
    raw_data = api.return_loan_orders(cur, 5)['offers'][0]
  File "/home/dog/poloniexlendingbot/modules/Poloniex.py", line 226, in return_loan_orders
    return self.api_query('returnLoanOrders', {"currency": currency, "limit": limit})
  File "/home/dog/poloniexlendingbot/modules/Poloniex.py", line 91, in api_query
    ret = urllib2.urlopen(urllib2.Request(req_url))
  File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/lib64/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 1258, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/lib64/python2.7/urllib2.py", line 1217, in do_open
    r = h.getresponse(buffering=True)
  File "/usr/lib64/python2.7/httplib.py", line 1089, in getresponse
    response.begin()
  File "/usr/lib64/python2.7/httplib.py", line 444, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python2.7/httplib.py", line 408, in _read_status
    raise BadStatusLine(line)
BadStatusLine:  Requesting returnLoanOrders
```

[...]

PR #239, opened 2017-02-21
Catch the bad status line exception and log it

laxdog:
Polo sometimes return a bad error code we can't parse. This catches that exception. 

## Description
Fixes https://github.com/Mikadily/poloniexlendingbot/issues/205

## TESTING STAGE
Tested

## Types of changes

## Checklist:

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -78,8 +78,6 @@ try:
         except KeyboardInterrupt:
             # allow existing the main bot loop
             raise
-        except BadStatusLine:
-            log.log_error("Caught BadStatusLine exception from Poloniex, ignoring.")
         except Exception as ex:
             log.log_error(ex.message)
             log.persistStatus()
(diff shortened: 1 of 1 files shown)


rnevet:
1. It's not so clear why this needs special handling. (to prevent sleep?)
2. I'm not sure the handling is complete without persisting the status. (i.e. log.persistStatus())

laxdog:
1) No reason. I actually thought they were separated because the first ones were catching the class and the others were using the string. 
2) I didn't realise that was required either. 

Changed and pushed.

rnevet:
Looks good to me, a bit of testing and it can be merged.

lendingbot.py
```diff
@@ -94,6 +95,8 @@
                 exit(1)
             elif 'timed out' in ex.message:
                 print "Timed out, will retry in " + str(Lending.get_sleep_time()) + "sec"
+            elif isinstance(ex, BadStatusLine):
```
chfast: Should it go as a `except BadStatusLine:`?

rnevet: It could but then we would need to duplicate the entire error logic that is in `except Exception:` already.

Issue #257, opened 2017-03-12
HTTP Error 422


Greveyaro:
![screen shot 2017-03-12 at 16 13 51](https://cloud.githubusercontent.com/assets/18435422/23851260/2c3c2e5e-07e3-11e7-9f64-b7cc9af9a67f.jpg)

laxdog:
I can't actually see your screen shot, I'm not sure if the upload failed or something. 

Can you just copy and paste the output from the terminal?

Greveyaro:
![screen shot 2017-03-13 at 17 50 10](https://cloud.githubusercontent.com/assets/18435422/23865316/92de5650-0815-11e7-983d-9229a909f937.jpg)

laxdog:
So the first one the 422, I haven't seen before, though it just looks to be problem connecting to Poloniex. It's not anything to worry about, but should probably be handled more gracefully by the bot. 

The second one (502: Bad Gateway) already has been fixed in the way I've mention, so you could do a pull on your repo to update and get the latest code.

rnevet:
I think we can handle all HTTP Errors without printing exceptions, error messages are enough.

laxdog:
Agreed, we can probably just change that line I put in the other day to look for 'HTTP Error' instead of 'HTTP Error 5'. 

I can raise a PR in the morning.

rnevet:
Or we can test for instance of httperror

laxdog:
Yip, OK.

rnevet:
Actually we can catch instance of URLError also covering #251

laxdog:
Opened https://github.com/Mikadily/poloniexlendingbot/pull/260

Missed this comment, so that's not in. I'll do it now though.

laxdog:
OK, both are in now.

Greveyaro:
![screen shot 2017-03-14 at 11 49 31](https://cloud.githubusercontent.com/assets/18435422/23897172/5733b7b6-08ac-11e7-9e17-5dd6eb6fef12.jpg)

Issue #251, opened 2017-03-08
Gets stuck in this error loop: Error: <urlopen error [Errno -5] No address associated with hostname> Requesting returnActiveLoans


taw00:
I have not put my finger on this yet, but I have to periodically restart lending bot (I use systemd) because it seems to have a network issue(?) and then never recovers(?).

I am not sure if other people are seeing this. I will continue to troubleshoot and see if I can determine what is going on. Maybe it is me, but it seems like that connectivity and recovery from connectivity issues is not robust.

By the way, love the tool in general.

```
Mar 08 13:04:17 rh python[27026]: 2017-03-08 13:04:17 Error: <urlopen error [Errno -5] No address associated with hostname> Requesting returnActiveLoans
Mar 08 13:04:17 rh python[27026]: Traceback (most recent call last):
Mar 08 13:04:17 rh python[27026]:   File "/home/taw/Apps/poloniexlendingbot/lendingbot.py", line 69, in <module>
Mar 08 13:04:17 rh python[27026]:     Data.update_conversion_rates(output_currency, json_output_enabled)
Mar 08 13:04:17 rh python[27026]:   File "/home/taw/Apps/poloniexlendingbot/modules/Data.py", line 83, in update_conversion_rates
Mar 08 13:04:17 rh python[27026]:     total_lended = get_total_lended()[0]
Mar 08 13:04:17 rh python[27026]:   File "/home/taw/Apps/poloniexlendingbot/modules/Data.py", line 45, in get_total_lended
Mar 08 13:04:17 rh python[27026]:     crypto_lended = api.return_active_loans()
Mar 08 13:04:17 rh python[27026]:   File "/home/taw/Apps/poloniexlendingbot/modules/Poloniex.py", line 159, in return_active_loans
Mar 08 13:04:17 rh python[27026]:     return self.api_query('returnActiveLoans')
Mar 08 13:04:17 rh python[27026]:   File "/home/taw/Apps/poloniexlendingbot/modules/Poloniex.py", line 104, in api_query
Mar 08 13:04:17 rh python[27026]:     ret = urllib2.urlopen(urllib2.Request('https://poloniex.com/tradingApi', post_data, headers))
Mar 08 13:04:17 rh python[27026]:   File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
Mar 08 13:04:17 rh python[27026]:     return opener.open(url, data, timeout)
[...]

PR #260, opened 2017-03-14
Catch all HTTP Errors

laxdog:
## Description

Handles more errors from Poloniex correctly, specifically HTTPError and URLError.

Fixes https://github.com/Mikadily/poloniexlendingbot/issues/251 and https://github.com/Mikadily/poloniexlendingbot/issues/257

## TESTING STAGE

Testing now

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -321,6 +321,9 @@ Notifications
 -------------
 The bot supports sending notifications for serveral different events on several different platforms. To enable notifications, you must first have a section in your config called ``[notifications]``, inside which you should enable at least one of the following events and also at least one notification platfom. The list of events you can notify about are:
 
+Notification events
+~~~~~~~~~~~~~~~~~~~
+
 - ``notify_tx_coins``
 
     - This will send a notification if any coins are transferred from your exchange account, to your lending account. You must have ``transferableCurrencies`` enabled for this to work.  Then you should set ``notify_tx_coins = True``.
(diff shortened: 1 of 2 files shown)


laxdog:
Don't merge this, there's something up. It's not catching what I expected. Need to test a bit more.

rnevet:
Not sure what issue you are facing, but HTTPError is a subclass of URLError if I'm not mistaken so you don't need to check for both.
https://docs.python.org/2/library/urllib2.html#urllib2.URLError

rnevet:
I'm thinking a bit slow on this but I guess we can handle PoloniexApiError the same way.

laxdog:
Yea, you're right:
```
>>> from urllib2 import HTTPError, URLError
>>> isinstance(HTTPError('', '', '', '', None), URLError)
True
```

The issue was that it didn't actually catch it. I put a pdb in after the last else to find out why, but it hasn't happened again yet.

laxdog:
Right finally found it again, looks like it's not catching it because it's on the notifiy_summary thread. So this patch does actually fix https://github.com/Mikadily/poloniexlendingbot/issues/257 , this is just another 422 that can happen on that thread. It'll need a separate patch.

I added the PoloniexApiError to catch and removed the HTTPError in favour of just using URLError. 

Everything should be up to date now.

PR #274, opened 2017-03-25
Send TCPServer shutdown signal from a separate thread

aaaronic:
This prevents the webserver from hanging on KeyboardInterrupt

## Description
When attempting to terminate the lending bot with the webserver running, the "Stopping WebServer" message printed and the process failed to terminate for many seconds, sometimes leading to a "Failed to stop WebServer:" message with a stack trace.

## TESTING STAGE
I'm open to suggestions of how to test this. It now exits cleanly for me, but I'm not sure how to add a good unit test for that.

## Types of changes
The source for socketserver has the following comment:
```
def shutdown(self):
    """Stops the serve_forever loop.

    Blocks until the loop has finished. This must be called while
    serve_forever() is running in another thread, or it will
    deadlock.
```
This change just follows through with that advice by sending the shutdown signal in a new thread specifically opened for this purpose.

## Checklist:

### DIFF
--- a/modules/WebServer.py
+++ b/modules/WebServer.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+import threading
 
 server = None
 web_server_ip = "0.0.0.0"
@@ -68,7 +69,7 @@ def start_web_server():
 def stop_web_server():
     try:
         print "Stopping WebServer"
-        server.shutdown()
+        threading.Thread(target = server.shutdown).start()
     except Exception as ex:
         ex.message = ex.message if ex.message else str(ex)
         print("Failed to stop WebServer: {0}".format(ex.message))


aaaronic:
This _might_ help address this todo in lendingbot.py: `TODO: Find out why non-daemon thread(s) are hanging on exit`

aaaronic:
I removed some of the original extraneous whitespace after opening the PR. My apologies.

laxdog:
The unit test haven't been merged to master yet, so I wouldn't worry about that. 

I'll test it myself later to verify, but it looks good to me. There's some more info here actually:

http://stackoverflow.com/questions/10085996/shutdown-socketserver-serve-forever-in-one-thread-python-application

I'm not sure if it's saying we should catch the KeyboardInterrupt at this level too, or it's OK the way it is. I'll have a better read when I test it.

laxdog:
Also, thanks for this :)

rnevet:
So if we're not waiting for the server to shutdown will this create an issue with restarting on the same port?

laxdog:
We still wait to shutdown the server, we just start another thread that can do it in the background and wait there as the main thread is about to die. 

The reuse / time_wait shouldn't cause any issues as the whole thing will just shutdown. In fact the reason we probably had to use the reuse port code in the first place was because we didn't have something like this to shutdown the thread properly, so it still had the socket open. So we can maybe remove that, though I don't think it's causing any harm.

rnevet:
@laxdog @aaaronic 
if I get a confirmation on running it for a period of time I'll merge.

aaaronic:
I've run it for ~1-4 hours at a time a several times on my branch. I don't leave any of my computers on 24/7, so I can't vouch for that period of time.

laxdog:
Yea, I've had it running for over 24 hour now without issue.

PR #276, opened 2017-03-28
Dockerizing application

ivan1993spb:
Creating dockerfile

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+*.idea
+default.cfg
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:2.7
+
+ADD . /app
+
+WORKDIR /app
+
+EXPOSE 8000
+
+CMD ["python", "lendingbot.py"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,11 @@ FROM python:2.7
 
 ADD . /app
 
+VOLUME /apt/market_data
+
 WORKDIR /app
+
+RUN pip install numpy
 
 EXPOSE 8000
 
--- a/modules/Configuration.py
+++ b/modules/Configuration.py
@@ -2,8 +2,9 @@
 from ConfigParser import SafeConfigParser
 import json
 from decimal import Decimal
+import os
 
-config = SafeConfigParser()
+config = SafeConfigParser(os.environ)
 Data = None
 FULL_LIST = ['STR', 'BTC', 'BTS', 'CLAM', 'DOGE', 'DASH', 'LTC', 'MAID', 'XMR', 'XRP', 'ETH', 'FCT']
 # This module is the middleman between the bot and a SafeConfigParser object, so that we can add extra functionality


laxdog:
Hey, thanks for contributing. This is a good idea, just a couple of points

1) Would it be better to share the dir through VOLUME rather than copying it with ADD? If it's copied then you need to rebuild the image every time you update your config. It also means it will break for a new checkout unless you change the config straight away (which also won't be created on first run)

2) It would be good to install numpy through pip on build too, it's used in the MarketAnalysis module. 

3) Could you add a few lines in the dock about using it? Something like:
```
docker build -t PLB .
docker run PLB # Creates settings file
vim default.cfg
docker run PLB
```
Or however it's best to run it.

ivan1993spb:
Hello! Thanks for good software!

1. Ok. I know right way
2. Ok
3. Ok. I will add topic `Using docker` to installation.rst

ivan1993spb:
@Mikadily @laxdog Could you create dockerhub account with build? After that i will create docs with for docker

ivan1993spb:
@Mikadily @laxdog 

If you decide to use quay hosting you can use badge like that

[Docker Repository on Quay](https://quay.io/repository/ivan1993spb/poloniexlendingbot)

ivan1993spb:
@Mikadily @laxdog To use docker correctly we should provide using environment variables
like here
http://stackoverflow.com/questions/26586801/configparser-and-string-interpolation-with-env-variable

```
[API]
apikey = %(API_KEY)s
```

Starting with
```
docker run --rm -v $(pwd)/default.cfg:/app/default.cfg -e API_KEY=my_api_key plb
```

laxdog:
Why can't we just use the config file instead of passing them through?

ivan1993spb:
@laxdog what do you mean? For docker using env vars is common practice. Without providing env vars docker image with poloniexlendingbot is useless.

ivan1993spb:
@laxdog 
>If you mount the default.cfg as a volume in the container, then it will be able to get the secret and api key from there.
>I just don't see the reason for adding the env vars, that's all. I'm not saying you shouldn't do it, I just don't understand why you are doing it.

You are right. So I think we should delete default.cfg add config.py which will main config file. config.py will define all configures what app needs with default values and checking environment variables. This way provides using poloniexlendingbot in docker. This way is the common practice. All projects that uses docker do that. For example see https://www.stavros.io/posts/how-deploy-django-docker/ django, or laravel in php: https://github.com/laravel/laravel/blob/master/config/database.php
We should pass api_key, secret and other configures into container as environment variables. For the most configures we should provide default values.

laxdog:
There's a previous attempt here:

https://github.com/Mikadily/poloniexlendingbot/compare/master...p-s-dev:master

I've asked the owner of that repo for some help.

p-s-dev:
Environment variables are great.  If I want to run the container in Amazon Container Service, it is much easier to pass the secrets in by env var rather than setup a volume and make the configuration file available on the host.  Using env variables also makes it easy to have a startup script file for each instance of a lendingbot, with different auth credentials.  I don't know python well enough to comment on the difference between default.cfg and config.py.  I hacked a script to build the default.cfg file from the default.cfg.example file when the image was built, and again on startup.  Generally passing values by env variable is just supplemental to the configuration file.  The precedence order might be: get values from file and replace any also defined by an env variable.  Also, the ability to make a volume for the configuration file is more of a "runtime" concern.  It seems optional to document the ability to use a volume-replaced configuration file as part of the image creation specification.  The user is welcome to use "-v" when doing "docker run".  That might not mean the volume needs to be in the Dockerfile.

Personally, I like having the configuration in the container image (as well as the container instance) so I defined ENV vars in the Docker build script and parsed the example config to a real config, which runs at image creation time.  "config in the container" allows the container to be self-sufficient and more 
easily "run anywhere".  Docker containers excel in unnoticeable build time, due to onioning filesystem, as long as the configuration file is added near the end of the Dockerfile.  It might be easy to build a new "throwaway" container for each time the config changes.

The MarketData should probably be stored on a volume, so it is persisted between containers.  "-v ~/polobotData:/lendingbot/market_data "

I built from alpine so the image would be as small as possible, ~272 MB.  A container that 
might be used for development would need more things.  It might be worth having different Dockerfiles for different types of users.

In my PR, I had one shell script to build the docker image and another shell script to run an instance of the image.  The run script tries to kill any existing similarly named containers, removes the container so there isn't a container naming conflict, and runs a new container.

laxdog:
OK, cool. Thanks @p-s-dev that helps. 

So what I'm taking from this is that the best approach would be to change the way we're reading the configs first, so they'll look for an env var and if not found, read it from the default.cfg. That's fine, I think that's useful anyway. 

Agreed on the market data part, I'd actually completely forgot about that. However ideally all the containers would be able to share the same data, one writes and everyone can read. That might be a phase two consideration though. 

Though does this mean it's better to ADD the whole git checkout to the container then?

Do we need a similar script to kill the instance? Shouldn't they be starting under different names so you can have differently configured instances running?

@ivan1993spb Do you want to make the config changes or do you need a hand? You can use os.getenv to check for the var

```
>>> os.getenv('iDONTexist', 'iDO')
'iDO'
```

I think it can just be added to each of the get() calls in the configuration module. So every time it does a look up for the config it'll check the env first. So this https://github.com/Mikadily/poloniexlendingbot/blob/master/modules/Configuration.py#L46 , would become this:

```
value = os.getenv(option, config.get(category, option))
```

I haven't tested any of that though. I could have missed something.

rnevet:
@laxdog since we merged the other docker PR, what about this one?

laxdog:
We can kill it I guess. 

I still think there are some good ideas in here, but the current implementation I think is how most people will use it.

Issue #278, opened 2017-03-29
Notification config value being ignored


kryogenic:
Hi, it seems that (at least using telegram notifications) the value for `notify_xday_threshold` is always considered to be True.

Also, to avoid spamming I think it would be much better if a notification was sent only if a loan offer was filled. Currently `notify_xday_threshold` sends a notification every time a loan offer is placed, and since the bot cancels and replaces all orders every minute or so, this can result in a lot of notifications being sent.

Granted, I am using a very large value for `gaptop` (20000%), so my bot almost always has an xday offer open.

laxdog:
Yip, you're completely right about the always True, in fact I think it's an bug for the others too. I'll have a look at it. 

I also agree about only notifying when the order is filled. I didn't add it as I knew it would have been a lot more complicated and I wasn't sure if anyone was going to use the notification features. I can have a look at adding it though.

kryogenic:
I would almost say there should instead be a separate config value that notifies each time an order is filled. That would be a feature request though so I'll leave it up to you whether you wanna embark on it. Thanks!

laxdog:
Yea, I reckon if I was going to do it I'd leave both. It's a feature request either way ;)

I'm just not sure exactly how to do it. I'll have to keep track of the active loans and check the time of the newest one against the last run time or something. There's potential to add it into some of the market analysis stuff, or at least leverage some of it.

kryogenic:
Ahh yeah I would want to connect to their websocket or something. I'm not sure but returnLendingHistory may also work. However it looks like the bot currently doesn't use either of those.

laxdog:
I think you can only get the public information from the web socket, the only way to get your loan stuff is through the REST API. 

returnLendingHistory will only show the completed loans, though that's something I'd like to add too.

kryogenic:
Ok I went ahead and partially implemented this, it's quite messy and just prints to console instead of actually sending the notifications but hopefully this gives you something to work off of!

All I did was add a little bit to the top of get_total_lends() in modules/Data.py:

`loans_provided = {}

def get_total_lended():
    global loans_provided
    crypto_lended = api.return_active_loans()
    if loans_provided == {}:
        loans_provided = crypto_lended['provided']
    else:
        new_loan_ids = set(map(lambda l: l['id'], crypto_lended['provided'])) - set(map(lambda l: l['id'], loans_provided))
        if new_loan_ids:
            print('new loans found: ', new_loan_ids)`

[...]

PR #281, opened 2017-03-29
Fix notify_xday_threshold, implement notify_new_loans option #278

kryogenic:
## Description
Details are in this issue: https://github.com/Mikadily/poloniexlendingbot/issues/278

## TESTING STAGE
Working so far!

## Types of changes

## Checklist:

### DIFF
--- a/modules/Lending.py
+++ b/modules/Lending.py
@@ -106,10 +106,9 @@ def notify_new_loans(sleep_time):
     global loans_provided
     new_provided = api.return_active_loans()['provided']
     if loans_provided:
-        get_ids = lambda loans: map(lambda l: l['id'], loans)
-        new_loan_ids = set(get_ids(new_provided)) - set(get_ids(loans_provided))
-        for id in new_loan_ids:
-            loan = filter(lambda l: l['id'] == id, new_provided)[0]
+        get_id_set = lambda loans: set([x['id'] for x in loans]) # lambda to return a set of ids from the api result
+        for id in get_id_set(new_provided) - get_id_set(loans_provided):
+            loan = [x for x in new_provided if x['id'] == id][0]
             t = "{0} {1} loan filled for {2} days at a rate of {3:.4f}%"
             text = t.format(loan['amount'], loan['currency'], loan['duration'], float(loan['rate']) * 100)
             log.notify(text, notify_conf)
--- a/modules/Lending.py
+++ b/modules/Lending.py
@@ -33,6 +33,7 @@ hide_coins = True
 coin_cfg_alerted = {}
 max_active_alerted = {}
 notify_conf = {}
+loans_provided = {}
 
 # limit of orders to request
 loanOrdersRequestLimit = {}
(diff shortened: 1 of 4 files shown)


modules/Lending.py
```diff
@@ -98,6 +102,21 @@ def notify_summary(sleep_time):
     scheduler.enter(sleep_time, 1, notify_summary, (sleep_time, ))
 
 
+def notify_new_loans(sleep_time):
+    global loans_provided
+    new_provided = api.return_active_loans()['provided']
+    if loans_provided:
+        get_ids = lambda loans: map(lambda l: l['id'], loans)
```
laxdog: I agree with Codacy here, I think this is more readable:
```
get_ids = lambda loans: [x['id'] for x in loans]
```

modules/Lending.py
```diff
@@ -98,6 +102,21 @@ def notify_summary(sleep_time):
...
+        new_loan_ids = set(get_ids(new_provided)) - set(get_ids(loans_provided))
+        for id in new_loan_ids:
+            loan = filter(lambda l: l['id'] == id, new_provided)[0]
```
laxdog: Up to you on this one
```
loan = [x for x in new_provided if x['id'] == id][0]
```

One thing though, what if there's more than one loan in your 60s sleep period?

kryogenic: List comprehensions ftw! I'd take both those changes, I'm just used to using callbacks and nodejs.

> One thing though, what if there's more than one loan in your 60s sleep period?

That's what the loop is for. The filter just gets the list of all loans with a specific ID, which should always have a single element.

modules/Lending.py
```diff
@@ -73,10 +74,13 @@ def init(cfg, api1, log1, data, maxtolend, dry_run1, analysis, notify_conf1):
 
     sleep_time = sleep_time_active  # Start with active mode
 
-    if notify_conf['enable_notifications'] and notify_conf['notify_summary_minutes']:
+    if notify_conf['enable_notifications']:
```
rnevet: This change will make the scheduler thread run even if no notifications are configured.

kryogenic: The thread will run but if no events are configured it will exit immediately.

kryogenic: In retrospect we don't actually need to check `if notify_conf['enable_notifications']:` at all. To make things more clear we could check `if scheduler.empty():` before creating and starting the thread.

kryogenic:
Ok, I've tested this merge req for 24 hours now, no bugs but I've noticed a couple oddities..

1. The notifications are far from instant, and are usually over a minute late. Strange cause the thread is supposed to run every 60 seconds. Perhaps this is due to 429 too many requests errors.

2. The scheduler thread will occasionally run into an exception and stop sending notifications. This happens more often with the patch, since we are polling `returnActiveLoans`

modules/Lending.py
rnevet: `if scheduler.empty(): `
Lets do that, keep it clear and clean.. (no body will get confused later).

rnevet:
Regarding 2, we had the same issue in the MarketAnalysis thread - i.e. network exceptions would kill it.

That's why the main loop is surrounded by a try clause:
https://github.com/Mikadily/poloniexlendingbot/blob/master/modules/MarketAnalysis.py#L53

Could maybe use a less harsh message.

modules/Lending.py
```diff
@@ -98,6 +102,20 @@ def notify_summary(sleep_time):
     scheduler.enter(sleep_time, 1, notify_summary, (sleep_time, ))
 
 
+def notify_new_loans(sleep_time):
+    global loans_provided
+    new_provided = api.return_active_loans()['provided']
+    if loans_provided:
+        get_id_set = lambda loans: set([x['id'] for x in loans]) # lambda to return a set of ids from the api result
+        for id in get_id_set(new_provided) - get_id_set(loans_provided):
```
rnevet: @kryogenic what about this codacy comment?

kryogenic:
Latest commit should cover everything.

The only problem I can see now is an increased number of `429 Too Many Requests` errors due to the excess calls to returnActiveLoans.

rnevet:
Are you working on a shared IP? The Poloniex API implementation should prevent single instance going over the call rate limit. (You can try reducing the rate from 6 per sec.)

rnevet:
Oh, maybe we need to make the API rate checking thread safe...

rnevet:
I'm not sure if pass on http errors might lead to not being able to find problems....

kryogenic:
It seems like the 429 were a result of Poloniex experiencing heavy traffic this week. I stopped getting them completely even with the notifications turned on.

For the http errors, it's up for discussion. I guess I should just follow the project conventions though and log that we're ignoring the error. Will fix.

laxdog:
Tested this for a while now and it looks good. Nice work, this is super useful.

rnevet:
Great, I'll merge once the exceptions are logged.

PR #246, opened 2017-02-28
Display Total Account Value

rnevet:
Display total account value (Lending, Margin, Exchange) in BTC and Output currency

## Description
Implements issue https://github.com/Mikadily/poloniexlendingbot/issues/240
Closes #240

## TESTING STAGE
In progress 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] **I have read CONTRIBUTING.md**
- [X] **I fully understand [Github Flow.](https://guides.github.com/introduction/flow/)**
- [x] **My code adheres to the [code style of this project.](https://poloniexlendingbot.readthedocs.io/en/latest/contributing.html)**
- [ ] **I have updated the documentation in /docs if I have changed the config, arguments, logic in how the bot works, or anything that understandably needs a documentation change.**
- [ ] **I have updated the config file accordingly if my change requires a new configuration setting or changes an existing one.**
- [X] **I have tested the bot with no issues for 24 continuous hours. If issues were experienced, they have been patched and tested again.**

PR #286, opened 2017-03-31
Web view improvements (update interval, settings)

ckcollab:
## Todo

- [x] Save settings to localStorage
- [x] Read settings from localStorage
- [x] Modal for settings
- [x] Gear icon
- [x] Add "years" to time views
- [x] ~~Limit USD view to only show 2 decimal places~~ Believe this is being handed off to Total Account Holdings feature? 

## Description

This makes the page update every 10 seconds instead of every 30. 

## TESTING STAGE

Currently testing this for 24hrs

## Types of changes

- [~] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -247,6 +247,8 @@ function Timespan(name, multiplier) {
             }
             if (currency == "BTC") {
                 return displayUnit.formatValue(earnings) + " <span class=" + currencyClass + ">" + displayUnit.name + "</span> / " + name + "<br/>"
+            } else if (currency == "USD" || currency == "USDT") {
+                return printFloat(earnings, 2) + " <span class=" + currencyClass + ">" + currency + "</span> / "+  name + "<br/>";
             } else {
                 return printFloat(earnings, 8) + " <span class=" + currencyClass + ">" + currency + "</span> / "+  name + "<br/>";
             }
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -214,18 +214,18 @@ function handleLocalFile(file) {
 function loadData() {
     if (localFile) {
         reader.readAsText(localFile, 'utf-8');
-        setTimeout('loadData()',30000)
+        setTimeout('loadData()', 10000)
     } else {
         // expect the botlog.json to be in the same folder on the webserver
         var file = 'botlog.json';
         $.getJSON(file, function (data) {
             updateJson(data);
             // reload every 30sec
-            setTimeout('loadData()',30000)
+            setTimeout('loadData()', 10000)
         }).fail( function(d, textStatus, error) {
             $('#status').text("getJSON failed, status: " + textStatus + ", error: "+error);
             // retry after 60sec
-            setTimeout('loadData()',60000)
+            setTimeout('loadData()', 60000)
         });;
     }
 }
--- a/www/lendingbot.js
+++ b/www/lendingbot.js
@@ -81,8 +81,6 @@ function updateRawValues(rawData){
         var maxToLend = parseFloat(rawData[currency]['maxToLend']);
         var highestBidBTC = parseFloat(rawData[currency]['highestBid']);
 
-        totalCoinsOverall += totalCoins
-
         if (currency == 'BTC') {
             // no bids for BTC provided by poloniex
             // this is added so BTC can be handled like other coins for conversions
@@ -186,6 +184,7 @@ function updateRawValues(rawData){
                 }
             }
         }
+        totalCoinsOverall += earningsOutputCoinRate * totalCoins * highestBidBTC;
     }
 
     // add headers
@@ -197,7 +196,7 @@ function updateRawValues(rawData){
     total_cell.innerHTML = "Total holdings";
     total_cell = total_row.appendChild(document.createElement("th"));
     total_cell.setAttribute("colspan", 2);
-    total_cell.innerHTML = totalCoinsOverall;
+    total_cell.innerHTML = prettyFloat(totalCoinsOverall, 2) + " USD";
 
     // show account summary
     if (currencies.length > 1 || summaryCoin != earningsOutputCoin) {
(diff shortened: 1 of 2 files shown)


rnevet:
Why? The bot has a slow cycle. How fast are you running the bot?

ckcollab:
Interval ~10seconds -- I love watching the log flying around right after I make changes, etc. across all my coins (and their prices). I'm used to streaming output so I like it as up-to-date as possible :)

I mean, ideally it should be configurable based on preferences. I could add a button/slider/drop down to set this value?

rnevet:
We have already some URL options.
It would be great to get a "settings" area if you can, otherwise another URL parameter should do it.

ckcollab:
Oh. I wasn't even aware of the URL options! I skimmed over that in the docs.

...maybe I can add a nice little section for managing all of these options... :)

For my reference http://poloniexlendingbot.readthedocs.io/en/latest/configuration.html#lendingbot-html-options

rnevet:
That would be great, never got to that. 
Keep in mind that the page is mobile friendly.

rnevet:
Just use prettyFloat instead of printFloat.

rnevet:
Actually I would like the time views to be selectable by checkboxes in the settings as well, personally I would look at daily, monthly only.

rnevet:
btw, I was working here to add "Total Account Value":
https://github.com/Mikadily/poloniexlendingbot/pull/246

www/lendingbot.js
```diff
@@ -246,6 +258,8 @@ function Timespan(name, multiplier) {
             }
             if (currency == "BTC") {
                 return displayUnit.formatValue(earnings) + " <span class=" + currencyClass + ">" + displayUnit.name + "</span> / " + name + "<br/>"
+            } else if (currency == "USD" || currency == "USDT") {
+                return printFloat(earnings, 2) + " <span class=" + currencyClass + ">" + currency + "</span> / "+  name + "<br/>";
```
rnevet: i.e. I use EUR. ;)

www/lendingbot.js
```diff
@@ -181,11 +184,20 @@ function updateRawValues(rawData){
...
     // add headers
     var thead = table.createTHead();
 
+    // total
+    var total_row = thead.insertRow(0);
+    var total_cell = total_row.appendChild(document.createElement("th"));
+    total_cell.innerHTML = "Total holdings";
+    total_cell = total_row.appendChild(document.createElement("th"));
+    total_cell.setAttribute("colspan", 2);
+    total_cell.innerHTML = prettyFloat(totalCoinsOverall, 2) + " USD";
```
rnevet: USD?

rnevet:
If you need a Gear icon you can use one from http://fontawesome.io/icons/

www/lendingbot.js
ckcollab: That would change behavior for quite a few coins, I think we want to see Satoshi level resolution most of the times?

www/lendingbot.js
ckcollab: Since you're doing this in another PR I'll pull this stuff

ckcollab:
> If you need a Gear icon you can use one from http://fontawesome.io/icons/

Just using the Bootstrap one, is fontawesome also included? Probably not necessary to have two icon libs when we only have the 1 page.

www/lendingbot.js
ckcollab: What other currencies should we add here? Some currencies may need special printing, in this case for USD it makes sense.

ckcollab: Also, the other currencies in this function use `printFloat`, you want to go in and change these to the style you prefer?

rnevet: @ckcollab "EUR" for example and actually any currency supported by https://blockchain.info/api/exchange_rates_api

I'm thinking about using the prettyFloat as it will provide an accuracy of 2 decimals for any coin and I think that should be good enough, no? or should we make the accuracy configurable? 

I'm open for suggestions.

ckcollab: Suggestion: I also add EUR, but maintain consistency with the surrounding code. Then in a separate PR we decide how to handle this particular problem.

rnevet: agreed.

rnevet:
@laxdog @Evanito can you take a look, IMO this is a great improvement for handling display settings.

ckcollab:
> Gear icon looks out of place and a little lopsided.

Goahead and replace, I don't mind :)

> Entering out of bounds values for refresh time are "saved" when an error should be shown.

Added a check for this.

> Would you be willing to move the other web settings to this settings box?

I'd rather not, very very busy! Glad to share what little I could given I use this all the time, thanks for all the hard work!

rnevet:
It is great! We will build upon. :)

Evanito:
I move to merge in its current state, since I don't want to muck around with multiple branches while trying to change a single icon.
Thank you for your contribution!

@rnevet I leave final merge to you.

rnevet:
@Evanito I've made the rest of the changes and minor fixes in this branch https://github.com/Mikadily/poloniexlendingbot/tree/web_setting. Made a PR back to @ckcollab and when he merges it, it should appear here.

rnevet:
@yura-pakhuchiy can you test this as well?

yura-pakhuchiy:
LGTM! I had exactly the same idea for settings dialog on webpage itself for quite some time. I have never found time to implement it. @ckcollab @rnevet Greak work!

Small things which can be improved:
1. Use 192 icon in navbar. It will look sharper on mobiles or if user zooms in on desktop.
2. Auto-collapse navbar after user clicked on settings. It is annoying a bit that navbar is still open after user closes settings dialog.

rnevet:
@ckcollab Thanks!

ckcollab:
@rnevet my pleasure, thanks for this project!

laxdog:
Just saw these changes as I've been working on an old branch. They look really great. Nice work @ckcollab and @rnevet

ckcollab:
Least I could do :)

PR #292, opened 2017-04-06
Market analysis macd

### LINKED ISSUE #267: MarketAnalysis.delete_old_data() takes too long to run
From my testing, it's adding about 7 seconds on in my env where I'm checking all currencies. I had 5 days of data for it to parse. 

It doesn't need to be run every time we write to disk, I don't think we can have a separate thread for it as locking would be an issue if both threads were updating the file. 

I don't actually think that having a rolling file is the best way to do it either, it would be better for the files to roll over each day.

I'd also like to change the long date stamp we have at the minute to epoch time. It will break backwards compatibility for those files people have (unless they convert them, or we add something to convert them), but it should speed up the reading and parsing of the file too for market analysis processing later.

### LINKED ISSUE #315: Bugs / required features for new market analysis module
Keeping a track of these here so I can work through them:

1) During the initial data collection interval the bot lends as if MA it turned off. This is undesirable for me (and probably most people), so needs a method to stop it and needs to be configurable
2) Still getting lots of `OperationalError: database is locked`. Full traceback:
```
Traceback (most recent call last):
  File "/home/dog/MACD_testing/modules/MarketAnalysis.py", line 91, in update_market_thread
    self.update_market(db_con, cur, self.recorded_levels)
  File "/home/dog/MACD_testing/modules/MarketAnalysis.py", line 122, in update_market
    db_con.execute(insert_sql)
```
3) Had a too many open files error, not sure how it was caused yet:
```
Error in MarketAnalysis: <urlopen error [Errno 24] Too many open files> Requesting returnLoanOrders
Traceback (most recent call last):
  File "/home/dog/MACD_testing/modules/MarketAnalysis.py", line 91, in update_market_thread
    self.update_market(db_con, cur, self.recorded_levels)
  File "/home/dog/MACD_testing/modules/MarketAnalysis.py", line 109, in update_market
    raw_data = self.api.return_loan_orders(cur, levels)['offers']
  File "/home/dog/MACD_testing/modules/Poloniex.py", line 242, in return_loan_orders
    return self.api_query('returnLoanOrders', {"currency": currency, "limit": limit})
  File "/home/dog/MACD_testing/modules/Poloniex.py", line 40, in new_method
    return method(self, *arg, **kws)
  File "/home/dog/MACD_testing/modules/Poloniex.py", line 108, in api_query
    ret = urllib2.urlopen(urllib2.Request(req_url))
  File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/lib64/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 1258, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/usr/lib64/python2.7/urllib2.py", line 1214, in do_open
    raise URLError(err)
```
4) Old data isn't being deleted (code for it locks the DB continually)
5) All config values are hardcoded
6) A historical source of rates would greatly improve start up speed. Anyone know of somewhere storing rates that we could query?
7) Current rate used is not weighted, so dust offers can still skew the price. I've a column in the DB for percentile, and old code to add to it. Just need to pull that in. 
8) Averages are calculated with pandas.rolling_mean, we don't need this any more and it's taking extra cycles. It can be a simple average calculated every interval.
9) All the rates used should be logged in the DB. Might make sense to do this with rolling_mean and do it less frequently.
10) The percentile calculation is no longer working as it's meant to. This will be partially fixed by deleting old data, but it should be given a fixed size array and / or resampled in pandas.

laxdog:
## Description

Most of the information for this is in the documentation. So please go there for details. 

These changes improve the recording of data by the market analysis module by:
1) Threading the writes per currency being recorded
2) Having more data in the files (more accurate time and top 10 offers)
3) Giving a more realistic (closer to the actual value) and deterministic (currencies reads don't block other currencies due to long waits as they're threaded) time between each row written.
4) Using MACD to work out the minimum rate to lend.

Closes #267 
Closes #315

## TESTING STAGE
In testing now

## Types of changes

## Checklist:


themouette:
I've been playing with poloniex lending feature recently, and it seems relevant to store the "total loan offers", as it can help a lot to predict how far in the offers book it will take to reach higher offers.

I explain:

When there is 100BTC offered, we are quite sure to reach a peak, as the latest offers are very high.
When there is 2kBTC in the offer book, the first 200BTC are following a linear (almost flat) progression.

IMHO, this is a very valuable information, because it helps predicting (and analysing) bursts. WDYT?

laxdog:
@themouette Do you mean the total amount being offered for the currency?

It would certainly be useful information to have, I'd just be worried about how long it would take to transmit as I don't think you can get that in an API call from Polo, we'd have to ask for the whole loan book and total it ourselves. I can do some testing though, as I do think it would be useful. 

The code in this PR is currently saving the top 10 loans, though that will be configurable. If it turns out it's not too bad to request the entire book, then we could look at saving that instead of just 10. It takes up a lot of data though.

themouette:
From my understanding, the user interface uses `https://poloniex.com/public.php?currency=BTC&command=returnLoanOrders&limit=999999` call to public api to determine the total amount.

It is very fast as you do not notice it being executed on every data refresh:

Taken from the HTML in lending page:

```
	    for (side in data){
			var lendingOfferTotal = 0;
			var offers = data[side];
			var html = '<table class="lendingOrderBook" id="lending' + side + 'Table"><thead><tr>';
			html += '<th>Rate</th>';
			html += '<th>Amount (' + lendingCurrency + ')</th>';
			html += '<th>Duration</th>';
			html += '</tr></thead><tbody>';
			for (x in offers){
				lendingOfferTotal += parseFloat(offers[x].amount);
				if (x > lendingOrderBookLimit)
					continue;
				
				rateString = (parseFloat(offers[x].rate) * 100).toFixed(4);
				amountString = parseFloat(offers[x].amount).toFixed(8);
				rangeString = parseInt(offers[x].rangeMin).toString();
				if (offers[x].rangeMax > offers[x].rangeMin)
					rangeString += "-" + parseInt(offers[x].rangeMax).toString();
				html += '<tr onclick=\'$(\"#lendingOfferRate\").val(\"' + rateString + '\");\'>';
				html += '<td>' + rateString + '%</td>';
				html += '<td>' + amountString + '</td>';
				html += '<td>' + rangeString + ' Days</td>';
				html += '</tr>';
			}
			if (offers.length > lendingOrderBookLimit)
				html += '<tr class="messageTR"><td></td><td></td><td class="messageTD"><a class="standard nounderline" onclick="renderLendingOrders(100);" href="javascript:void(0);">Load More<i class="fa fa-sort-amount-asc"></i></a></td></tr>';
			html += "</tbody></table>";
			
			$("#" + side + "Total").html(lendingOfferTotal.toFixed(8));
			$("#lending" + side).html(html);
			$("#lending" + side + 'Table').dataTable({
			    'paging': false,
			    'autoWidth': true,
			    'info': false,
			    'scrollY': 198,
			    'bSort' : false,
				'language': { "emptyTable": "There are no loan " + side + "." },     
			    'fnInitComplete': function() {
			        $('#lending' + side + 'Table_wrapper')
			        .find('.dataTables_scrollBody')
			        .jScrollPane({
			            verticalDragMinHeight: 20
			       });
			    }          
			});
			$("#lending" + side + 'Table').DataTable().draw();
			if (loadMore)
				$('#lending' + side + 'Table_wrapper').find('.dataTables_scrollBody').data('jsp').scrollToY(9999999);
		}
```

laxdog:
Yea, that's the request that's required. 

I did some testing, and on my machine, requesting 10 orders takes about 250ms, requesting the max available (around 1000 at the time) took about 360ms. 

So there is a noticeable difference, though I don't think it's enough to worry about. 

I'll add a column to the table I'm recording with total volume then. 

It could also be a good measure for the removal of dust offers, I'm currently implementing it via a percentile of the 10 best offers. I'll have a think anyway.

laxdog:
This is at the point it could be tested if people were feeling brave, I've a lot more to do though. Everything is hardcoded and there are no docs. Also, it probably only works on linux / mac until I add options for where to store the database.

modules/MarketAnalysis.py
```diff
@@ -148,3 +196,84 @@ def get_percentile(self, rates, lending_style, use_numpy=use_numpy):
...
+            sys.stdout.write("Short higher : ")
+            rate = short_rate if rates_df.iloc[-1][1] < short_rate else rates_df.iloc[-1][1]
+        else:
+            sys.stdout.write("Long  higher : ")
+            rate = long_rate
+        # TODO Needs config option
+        rate = rate * 1.05
+        return rate
+
+    def create_connection(self, db_path='/tmp/plb.db', db_type='sqlite3'):
```
rnevet: We should use a relative path...
http://stackoverflow.com/a/5137509/348378

laxdog: Yea, OK. Good idea. 

I'll put it in with my next commit. Wasn't really expecting anyone to look at it yet, so I didn't bother with a lot of stuff :)

modules/Lending.py
```diff
@@ -201,6 +201,7 @@ def cancel_all():
 
 
 def lend_all():
+    print("Rate : {0}".format(Analysis.get_rate_suggestion('BTC')))
```
Evanito: Leftover from debug?

modules/MarketAnalysis.py
```diff
@@ -25,79 +41,131 @@ def __init__(self, config, api):
...
+            print("Error in MarketAnalysis: {0}".format(ex.message))
+            traceback.print_exc()
+
+    def delete_old_data_thread(self, cur):
+        time_in_sec = 1800  # TODO Take this from config
         while True:
             try:
-                self.update_markets()
-                self.delete_old_data()
+                db_con = self.create_connection(cur)
```
Evanito: What is the significance of con in db_con? I'm curious

modules/MarketAnalysis.py
```diff
@@ -106,9 +174,17 @@ def get_rate_suggestion(self, cur, rates=None, method='percentile'):
             if len(rates) == 0:
                 return 0
             if method == 'percentile':
+                # rates is a tuple with the first entry being unixtime
+                rates = [x[1] for x in rates]
                 return self.get_percentile(rates, self.lending_style)
             elif method == 'golden_cross':
-                raise ValueError("{0} not yet implmented")
+                if len(rates) < 1800:
```
Evanito: In this case, consider a fallback to percentile?
Probably doesn't matter either way.

modules/MarketAnalysis.py
```diff
@@ -148,3 +225,87 @@ def get_percentile(self, rates, lending_style, use_numpy=use_numpy):
...
         result = truncate(result, 6)
         return result
+
+    def get_golden_cross_rate(self, cur, rates_df, short_period=150, long_period=1800):
+        # TODO These don't need to be rolling mean, simple mean on the last X will do (and will be faster)
+        short_rate = pd.rolling_mean(rates_df[1], window=short_period, min_periods=1).iloc[-1]
+        long_rate = pd.rolling_mean(rates_df[1], window=long_period, min_periods=1).iloc[-1]
+        # TODO remove the sys writes
+        if short_rate > long_rate:
+            sys.stdout.write("Short higher : ")
```
Evanito: Spammy

modules/MarketAnalysis.py
```diff
@@ -148,3 +225,87 @@ def get_percentile(self, rates, lending_style, use_numpy=use_numpy):
...
+            rate = short_rate if rates_df.iloc[-1][1] < short_rate else rates_df.iloc[-1][1]
+        else:
+            sys.stdout.write("Long  higher : ")
+            rate = long_rate
+        # TODO Needs config option
+        rate = rate * 1.05
```
Evanito: Why?

modules/Lending.py
laxdog: Yea, there's probably a few more in there. It's useful to get the rates that out of the module without actually making a loan. 

They'll probably stay until it's suitable for merging to master. I should probably put 'do not merge' on this PR actually.

modules/MarketAnalysis.py
laxdog: It's the connection to the sqlite database created on disk. In this implementation there's a db per currency as I was trying to increase concurrency. I'm still not sure that's the best way to do it though as I'm still having issues deleting and sometimes (though less frequently) on writes.

modules/MarketAnalysis.py
laxdog: The get_rate_suggesion method takes a method argument now, so if we want we can do that functionality there. Meaning we could catch the correct exceptions and run a different method if it makes sense. 

The default for the method is golden_cross, which I'd actually like to keep, but it will check for pandas / numpy and act accordingly.

modules/MarketAnalysis.py
laxdog: Yea, this is all debug code.

modules/MarketAnalysis.py
laxdog: The * 1.05?

The assumption here is that if you have a short rate that is rising above the long rate then the market is in a period where it's rising. So here when we return the rate we're trying to catch the top of the spike. It's not the best way to do it, but it works. 

I'd initially thought making this configurable would be a good way to implement this. Now though after chatting to @rnevet I think we both agree that this module should feed into the gapbottom and gaptop (I actually want to be able to drive those AND cur_daily_min via config from the MA module, possibly even with different methods, though that's long term).

So if I do it with the top/bottom method this will probably change. Here I was trying to get the best value for just one loan. When I was developing this I wasn't using spread at all and I was getting really good results. Though it's clear you can get even better results using the spread if it's clever enough.

In the new top/bottom method I'm going to take the trend of the short term moving average and extrapolate the plot to work out where the trend looks like it's moving to. The example I gave was this:

```TIME - Short - Long
12:00 - 0.08 - 0.1
12:01 - 0.09 - 0.1
12:02 - 0.1 - 0.1
12:03 - 0.11 - 0.1
12:04 - 0.12 - 0.1
```
In that example, the gapbottom would be 0.12 and gaptop would be 0.14. So we'd be making the assumption that the current upward trend continues.


I hope you meant the * 1.05, otherwise I've just gone on a massive rant! :)

modules/MarketAnalysis.py
```diff
@@ -16,105 +20,236 @@
...
-
-    def delete_old_data(self):
-        for cur in self.open_files:
-            with open(self.open_files[cur], 'rb') as file_a:
-                new_a_buf = StringIO()
-                writer = csv.writer(new_a_buf)
-                reader2 = csv.reader(file_a)
-                for row in reader2:
-                    if self.get_day_difference(row[0]) < self.max_age:
-                        writer.writerow(row)
```
Evanito: else: 
    break

I am foolish for not seeing this waste earlier.

modules/MarketAnalysis.py
```diff
@@ -16,105 +20,237 @@
...
+
+        :return: A float with the suggested rate for the currency.
+        """
+        if method == 'percentile':
+            seconds = self.percentile_seconds
+        elif method == 'golden_cross':
+            seconds = self.MACD_long_win_seconds
+        else:
+            print("Error: {0} method not recognised, falling back to percentile")
+            method == 'percentile'
```
Evanito: Needs addressing

themouette:
from what I can see, someone is manipulating rates to keep them really low, It looks like this MR (and some strategy when not using market analysis) to address this issue.

For instance, this morning, there were less than 3k BTC offered for lending, and lending rate was around 0.1%.

Few weeks ago, it would have been at least 0.4 ou 0.5.

laxdog:
I was going to ask for review, but I see there are a lot of codebeat issues.

laxdog:
That's going to take me a while. @rnevet and @Evanito could you start having look at this and I'll work my way through those issues.

rnevet:
We just added codebeat, not sure if it isn't too strict, don't go too crazy with fixing the problems reported by it. Just go through them see if things make sense to you. Can highlight which checks seem too strict.

laxdog:
Some I think are going to be too hard to fit in with it's requirements. Though it's quite interesting analysis in general.

modules/MarketAnalysis.py
```diff
@@ -148,3 +292,95 @@ def get_percentile(self, rates, lending_style, use_numpy=use_numpy):
...
+        :param long_period: Length in seconds of the long window for MACD calculations
+        :param multiplier: The multiplier to apply to the rate before returning.
+
+        :retrun: A float of the suggested, calculated rate
+        """
+        if multiplier is None:
+            multiplier = self.daily_min_multiplier
+        short_rate = rates_df.rate0.tail(short_period).mean()
+        long_rate = rates_df.rate0.tail(long_period).mean()
+        # TODO remove the sys writes
```
Evanito: Don't forget about this todo ;)

docs/market_analysis.rst
```diff
@@ -0,0 +1,291 @@
...
+
+The most important is probably the hidecoins change to False. This means that it will always place loans so you don't need to have as low a resolution on the sleep timers. You also want to make sure the gaptop and gapbottom are high so you can get a large spread.
+
+======================= =========
+Config                  Value
+======================= =========
+sleeptimeactive         10
+sleeptimeinactive       10
+spreadlend              3
+gapbottom               400
```
rnevet: Consider new raw gaps mode.

druilaap:
I like the multi-threaded market data recording. Is there anyway I can help to get this merged? Is it worth it for me to resolve the Codebeat warnings?

rnevet:
@druilaap You can see the comments I left and sync with @laxdog if he needs help... Codebeat is considered nice to have, wouldn't start there.

laxdog:
OK, I'm going to look at all these issues soon. Just needed a bit of a break :)

I'll sort it out @druilaap but thanks for the offer.

laxdog:
I'll find a way to test this Codebeat stuff locally, I'll be here all day pushing changes to test.

rnevet:
@laxdog did you see my comments?

laxdog:
The don't start with Codebeat one @rnevet ?

I thought it wouldn't take me long to be honest. Or you mean the other stuff? I've not read through it all yet.

rnevet:
lol, now that you mention it... the codebeat comment was also relevant. ;)  
But I meant the other ones.

docs/market_analysis.rst
```diff
@@ -0,0 +1,277 @@
...
+
+The module has two main methods to calculate the minimum rate:
+
+Percentile
+``````````
+This method takes all the data for the given time period (`percentile_seconds`_) and works out the Xth percentile (`lendingStyle`_) for that set of data. For example if you are using a ``lendingStyle`` of 85 and you had a list of rates like so
+
+  :Example: 0.04, 0.04, 0.05, 0.05, 0.05, 0.05, 0.06, 0.06, 0.06, 0.07, 0.07, 0.07, 0.08, 0.08, 0.09, 0.09, 0.09, 0.10, 0.10, 0.10
+
+The 85th percentile would be 0.985 because 85% of rates are below this. The following configuration options should be considered when using the percentile calculation method:-
```
laxdog: So I'm not sure what either of you are getting at here. 

This was just an example of how the percentile calculation works (for those who don't know). I haven't gone into the details of which DB etc it uses. This is still the old analysis method we had. It just uses the DB for the data now.

docs/market_analysis.rst
```diff
@@ -0,0 +1,291 @@
...
+`lendingStyle`_           The percentage used for the percentile calculation
+`percentile_seconds`_     The number of seconds to analyse when working out the percentile
+`MACD_long_win_seconds`_  The number of seconds to used for the long moving average
+`MACD_short_win_seconds`_ The number of seconds to used for the short moving average
+`keep_history_seconds`_   The age (in seconds) of the oldest data you wish to keep in the DB
+`recorded_levels`_        The depth of the lending book to record in the DB, i.e. how many unfilled loans
+`data_tolerance`_         The percentage of data that can be ignore as missing for the time requested in
+                          ``percentile_seconds`` and ``MACD_long_win_seconds``
+`daily_min_method`_       Which method (MACD or percentile) to use for the daily min calculation
+`MACD_multiplier`_        Only valid for MACD method. The figure to scale up the returned rate value from the MACD calculation
```
laxdog: Yes, I agree. But I'd really like to do that all in a big config fix that's really overdue. 

https://github.com/BitBotFactory/poloniexlendingbot/issues/364

docs/market_analysis.rst
laxdog: I'm running some tests now to get a better value for this, then I'll move it to RawBTC.

docs/market_analysis.rst
rnevet: Fine. Just making sure.

utdrmac:
I haven't been following along completely, but have you looked at this library:

https://tulipindicators.org/
https://tulipindicators.org/macd

https://github.com/cirla/tulipy

laxdog:
@utdrmac I haven't looked at that one in particular, but did have a look at talib, which I've used before and does similar stuff. I'd definitely like to add more analysis and have a load of ideas for things that can be added. But at the minute, I just want to get this merged. 

Once this is in it should be easier to add to and improve.

docs/market_analysis.rst
utdrmac: "..to used" -> "..to use"

utdrmac:
Has there been any backtesting on how much "improvement" MACD analysis will give vs the current algorithm? I guess I'm curious why MACD would be better than the current. Are we expecting better rate calculations? Better in what sense?

druilaap:
I am trying to write some code that sets the lending length based on past market data. At the moment this PR adds code that creates a database for each currency that is being tracked. Would it not be easier if all the market data is stored in one single database? That way, it would be much simpler to do sql queries like "When was the last time that any currency had a lending rate above 1%?". 
Sqlite3 is serialized by default, so it should be threadsafe.

PR #296, opened 2017-04-12
Avoid rate aberations in market analysis

themouette:
## Description

Current implementation always take the lowest rate in offers.

There is often some "aberations" data before the pack of offers with a
similar rate.

This commit removes the offers that are bellow the 5th percentile of the
30 first offers.

Consider the following offers:

- 0.00435
- 0.00597
- 0.00597
- 0.00597
- 0.00598
- 0.00599
- 0.00599
- 0.00599
...

The goal is to remove `0.00435`.

By not doing so, it takes the whole Poloniex rates lower.
## TESTING STAGE
Started qpril 12th

## Types of changes

## Checklist:


rnevet:
@laxdog how does this relate to the work you are doing?

I think I would prefer post processing while recording more information, not sure if that's not too "resource heavy"

laxdog:
@rnevet this is something I'm addressing, though not in this manner. As you say, it needs to be done after the results are recorded. We definitely should not be altering any information from Poloniex. 

I have added extra information to the market analysis output and I'm currently working on a suitable method for a weighted price to work from, using both the rate and the amount (as the weighting). 

I agree with the issue you're trying to fix @themouette , and for a quick fix this is perfect, but I think long term it's not a good idea. 

The future goal would be to allow people to write their own strategies, meaning we need to keep the information exactly as is.

Most of the work should be done by the end of next week. 

The changes I'm working on are in this PR (I've code locally not pushed) https://github.com/Mikadily/poloniexlendingbot/pull/292

themouette:
I agree this should be done afterward, and your work sounds great.

with current implementation there is a single stored value, so it was the simplest way to do it. I close the merge request in favor of #292

laxdog:
Been quite held up on the work in question, it's still on the way though :)

PR #298, opened 2017-04-15
Synchronized Poloniex API Calls

rnevet:
I've noticed that the 422 errors are related to making multiple calls in parallel to Poloniex API.
Additionally the limit_request_rate has an issue when used from multiple threads as it would only sleep the current thread.

Both issues are fixed and unit tested, synchronizing the calls might slow down the bot in some situations.

## TESTING STAGE
In progress.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_PoloniexAPI.py
+++ b/tests/test_PoloniexAPI.py
@@ -1,0 +1,47 @@
+import time
+
+# Hack to get relative imports - probably need to fix the dir structure instead but we need this at the minute for
+# pytest to work
+import os, sys, inspect
+currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parentdir = os.path.dirname(currentdir)
+sys.path.insert(0, parentdir)
+
+from modules.Poloniex import Poloniex
+import modules.Configuration as Config
+import modules.Data as Data
+import threading
+
+Config.init('default.cfg', Data)
+api = Poloniex(Config.get("API", "apikey", None), Config.get("API", "secret", None))
+
+
+def multiple_api_queries(n):
+    try:
+        for i in xrange(n):
+            # print 'api_query ' + str(i + 1)
+            thread1 = threading.Thread(target=api.return_open_loan_offers)
+            thread1.start()
+    except Exception as e:
+        assert False, 'api_query ' + str(i + 1) + ':' + e.message
+
+
+# Test fast api calls
+# def test_multiple_calls():
+    # multiple_api_queries(9)
+
+
+def api_rate_limit(n, start):
+    api.limit_request_rate()
+    # verify that the (N % 6) th request is delayed by (N / 6) sec from the start time
+    if n != 0 and n % 6 == 0:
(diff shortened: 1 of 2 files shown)


tests/test_PoloniexAPI.py
```diff
@@ -0,0 +1,47 @@
+import time
+
+# Hack to get relative imports - probably need to fix the dir structure instead but we need this at the minute for
+# pytest to work
+import os, sys, inspect
```
ckcollab: https://pypi.python.org/pypi/pytest-pythonpath

This may help?

rnevet: @laxdog

rnevet:
@laxdog @ckcollab @Evanito @yura-pakhuchiy I would like someone except me to test this before I merge. I have tested it for some days now with no apparent side effects.

ckcollab:
@rnevet on it. Pulled this branch and am testing now. Seems to be working fine from 30 seconds of testing :)

tests/test_PoloniexAPI.py
ckcollab: Maybe I could in a new PR make this small fix, for shits and giggles ;)

dt2017:
pulled and running on test machine now

tests/test_PoloniexAPI.py
rnevet: @ckcollab currently there are only 2 test files, it would be great if you can fix it. :)

ckcollab:
Still seems to be working fine for me.

rnevet:
@ckcollab thanks. @dt2017 if everything is fine as well, I'll merge this.

dt2017:
Belive so, and working well with the market data as well

rnevet:
@dt2017 Thanks!

tests/test_PoloniexAPI.py
laxdog: @rnevet and @ckcollab I really think we should fix the directory structure to get that working properly. 

I'd like to eventually push to PyPi and also if we don't need to add extra plugins, I don't think we should. 

It should really look more like this:

```
├── LICENSE
├── README.md
├── default.cfg.example
├── docs
│   └── docs
├── lendingbot.py
├── poloniexlendingbot
│   ├── market_data
│   │   └── readme.txt
│   ├── modules
│   │   └── modules
│   ├── tests
│   │   └── tests
│   └── www
│       ├── images
│       │   └── images
│       └── www
├── requirements.txt
└── setup.py
```
By all means have a look at that though if you like. I haven't gotten around to it. That's probably not the exact structure we should be using, but it's not far off. The important part is that we have a single folder that can be installed with everything needed to run inside it, in this case the poloniexlendingbot dir.

rnevet: @laxdog, that looks good, should we start an issue for discussion?

laxdog: Agreed @rnevet , opened here : https://github.com/Mikadily/poloniexlendingbot/issues/304

PR #306, opened 2017-04-25
Settings improvements

yura-pakhuchiy:
- Move select botlog.json button to settings dialog
- Clicking on label selects control in settings
- Cleanups

## Types of changes


rnevet:
@yura-pakhuchiy the only thing that bothers me is that it is now completely not intuitive to find the select file function. 
How about making it a button like settings (that will only be visible when needed) with a modal dialog?

yura-pakhuchiy:
@rnevet Actually selecting file in preferences seems quite intuitive for me, but your idea with new menu item sounds better! Developers might want change log file often, opening settings for this maybe annoying.

yura-pakhuchiy:
@rnevet I pushed commit which implements this

rnevet:
Very nice! 👍

PR #343, opened 2017-05-17
Add Raw gap setting

### LINKED ISSUE #269: Feature request gapbottom and gaptop by nominal units count, not by % of the portfolio
gapbottom and gaptop variable are defined how far into the lending book (in percent of the balance for the respective coin) the bot will go

the % of the balance is not so relevant, as my balance can be changed from time to time, and as I think that the strategy of where to place my lending order has nothing to do with my personal balance.

I prefer to always place my order in depth of 60 to 80 btc inside the lending book, 1000 to 1200 eth inside the lending book, etc…

Can you please add variables which are similar to gapbottom and gaptop, but defined by nominal unit numbers, not % of the balance? That way, I don't need to change the config file every time I withdraw or deposit money to/from the lending account.

In the example I gave, If I want depth of 60-80 btc for lending bitcoin and depth of 1000-1200 eth for lending ether, I should config the suggested 

nominal_gapbottom 	= 60,
nominal_gaptop		= 80,

and 1000, 1200 in the eth respected coinconfig fields.

Of course, it is recommended to add boolean variable to define if the bot will use the % gapbottom, gaptop variables or the nomninal variable.

Evanito:
Allows users to set a hard value to skip in the lendbook, as opposed to only offering values relative to their balance.

Has a `gapmode` setting that allows switching between `relative`, `raw`,  or `rawbtc` gaps (and possibly even add a setting that is relative to the market's volume?)

Closes #269 

## Types of changes

## Checklist:

### DIFF
--- a/default.cfg.example
+++ b/default.cfg.example
@@ -27,6 +27,9 @@ spreadlend = 3
 #before placing the first (gapbottom) and last (gaptop) offer.
 #If gapbottom is set to 0, the first offer will be at the lowest possible rate.
 #However some low value is recommended (say 10%) to skip dust offers.
+# Picking a raw value (in BTC) will override the other calculated values. Comment them out to return to normal.
+#rawgapbottom = 0.5
+#rawgaptop = 1
 gapbottom = 10
 gaptop = 200
 
--- a/default.cfg.example
+++ b/default.cfg.example
@@ -26,10 +26,9 @@ spreadlend = 3
 #The depth of lendbook (in percent of lendable balance) to move through
 #before placing the first (gapbottom) and last (gaptop) offer.
 #If gapbottom is set to 0, the first offer will be at the lowest possible rate.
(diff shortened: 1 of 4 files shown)


rnevet:
Would it make sense to have this option only in the specific coin config? it will then override the default behavior?

Evanito:
I was thinking that this is a change for default behavior, but per-coin settings could be added later in the coincfg cleanup/removal/overhaul I was planning.

rnevet:
I was actually thinking about making it simpler by having it only available as a specific coin config - not as a general config option.

Evanito:
The same code is pointed at by both config types:

The default value is set by the general setting, which can be overwritten for specific coins. Everything after that is the same.

Evanito:
Didn't want to do too much in one PR, coincfg cleanup is next.

PR #351, opened 2017-05-23
Fixed refreshstatus

Casimir1904:
Webif log fixed
## Description
Webif log didn't work.


laxdog:
There's no change here @Casimir1904 , it's only a whitespace difference. 

What are you changing?

laxdog:
Nothing changed.

PR #377, opened 2017-06-09
Configurable Poloniex API method URLs

riv1:
## Description
I've moved the Poloniex API urls to the config file. Just makes them easier to maintain going forward. Updated the configuration docs.

## TESTING STAGE
Regenerate the config from the default example, since the default values are stored in default.cfg.example. Tested for a few hours.

## Types of changes

- [x] 

## Checklist:

- [x ] **I have read CONTRIBUTING.md**
- [ ] **I fully understand [Github Flow.](https://guides.github.com/introduction/flow/)**
- [x ] **My code adheres to the [code style of this project.](https://poloniexlendingbot.readthedocs.io/en/latest/contributing.html)**
- [x ] **I have updated the documentation in /docs if I have changed the config, arguments, logic in how the bot works, or anything that understandably needs a documentation change.**
- [ x] **I have updated the config file accordingly if my change requires a new configuration setting or changes an existing one.**


riv1:
needs more testing for now.

riv1:
updated and ran tests

rnevet:
I actually don't see a good reason to do this, config is already cluttered.

riv1:
Its just good practice, I would not deem this as clutter.

rnevet:
@riv1 I didn't mean to offend of course.
Consider the following:
1. We might want to support Bitfinex (there is an open issue for that), so keeping the endpoint encapsulated in the API module makes sense.
2. We have many no-techy users

If Poloniex would give people different access points to their service I would find that this change make complete sense.

riv1:
@rnevet no offense taken. I agree on keeping things simple for end users. Also agree on encapsulating the endpoint. Rather define the access points as variables within the module, especially if more api calls are to be added in future.

rnevet:
I agree, it would be better to define the access points as variables within the module.

PR #388, opened 2017-06-13
added sleep in case of ip time ban (newly introduced by poloniex)

JediWed:
Poloniex changed it's api due high load on their server.
When someone makes too many requests in a short time period the IP will be banned for 2 minutes.
poloniex recommends to sleep 130 seconds

## Description
check for http error 429.
if it happens an error message is printed and a sleep for 130 seconds starts

## TESTING STAGE
tested, sleep works properly in case of a ban
just restart bot several times in a short timeslot to get banned

## Types of changes

## Checklist:
- [N/A] **I have tested the bot with no issues for 24 continuous hours. If issues were experienced, they have been patched and tested again.**

24 h test not needed. A ban can be produced by restarting the bot few times in a short time frame to get banned. This will prove the functionality of the change.


rnevet:
There is a sleep in case an error is caught, should we calculated that it?

JediWed:
Set 130 hard because it's recommended by poloniex.
A better way would be maybe to parse the message from poloniex and use their recommendation but that could cause unforseen consequences if poloniex changes the error message.

rnevet:
Ah, and you don't need another PR, the firstone was updated since you pushed to the same branch.

JediWed:
Oh damn, sorry for that.
I will close that one here.

PR #416, opened 2017-07-03
Update gaps in example

rnevet:
Update gaps to a more reasonable value, using the RawBTC method.


rnevet:
@laxdog what do you think? simple reasonable values?

laxdog:
Yea, I need to pull these into my stuff as well and test it actually. I can't comment on the actual value as I not played with that code much. Seems fine though.

PR #417, opened 2017-07-04
Added support to multi template webserver trought config file

LeaderGL:
## Description

Added support to multiple templates for the webserver trought the configuration file.
User can select the target template setting the 'customWebServerTemplate' field.

## Types of changes

## Checklist:

- [N/A] **I have read CONTRIBUTING.md**
- [N/A] **I fully understand [Github Flow.](https://guides.github.com/introduction/flow/)**


rnevet:
Hi,
So if I understand correctly this should allow you to run different HTML/JS templates for presenting the bot status?  Do you have an example other then default already?
Could the default be the /www and not /www/default? as this change breaks current config as far as I can see.

LeaderGL:
rvenet yes this is the reason why. 
As test templates i've used the default one and the one, under pull request, of tamaskan (with some changes).
The default value can be whatever the user want, i use /www/default in order to have a template tree and you're right this broke the current config. I could use /www as default but this is a little bit dirty, any other suggestion?

LeaderGL:
updated code in order to no break old configurations

www/manifest.json
```diff
@@ -1,19 +1,19 @@
-{
```
rnevet: Why are there changes to this file?

LeaderGL: this is very strange, i've only "moved" back all the files from /www/default to /www without changing them. furthermore no ones of the other files in the same directory looks changed

LeaderGL:
documentation file updated

www/manifest.json
LeaderGL: also the commit log show the file as renamed without changes (https://github.com/BitBotFactory/poloniexlendingbot/pull/417/commits/ec485cc0f9ee86b4af3cf43c7bdf48db24074dc5)

modules/WebServer.py
```diff
@@ -4,29 +4,47 @@
...
-def initialize_web_server(config):
-    import threading
-    global web_server_ip, web_server_port
-
-    if config.has_option('BOT', 'customWebServerAddress'):
-        custom_web_server_address = (config.get('BOT', 'customWebServerAddress').split(':'))
-        if len(custom_web_server_address) == 1:
-            custom_web_server_address.append("8000")
-            print "WARNING: Please specify a port for the webserver in the form IP:PORT, default port 8000 used."
+def get_web_server_setting(config, option, default):
```
rnevet: config.get already has a default value option this function is not required:
https://github.com/BitBotFactory/poloniexlendingbot/blob/master/modules/Configuration.py#L44

modules/WebServer.py
```diff
@@ -4,29 +4,47 @@
...
+
+def initialize_web_server(config):
+    '''
+    Setup the web server, retrieving the configuration parameters
+    and starting the web server thread
+    '''
+    global web_server_ip, web_server_port, web_server_template
+
+    # Check for custom web server address
+    web_server_ip = get_web_server_setting(config, 'customWebServerAddress', '0.0.0.0')
```
rnevet: Can we find an elegant way not to break current config, or at least issue an error if web_server_ip contains the port information as it is in the current config?

www/manifest.json
rnevet: maybe whitespace difference, if you compare with master do you get a difference?

LeaderGL: no differences

LeaderGL:
webserver module updated with code review suggestions

rnevet:
@laxdog @Evanito any objections to merge?

laxdog:
Looks good.

PR #428, opened 2017-07-25
Add a Dockerfile

mahiso:
PR adds a Dockerfile to build docker image on (my) NAS or Docker hub.
No changes in other code, compose file or configuration.

## Types of changes

## Checklist:

- [ ] **My code adheres to the [code style of this project.](https://poloniexlendingbot.readthedocs.io/en/latest/contributing.html)**


rnevet:
@mahiso I'm missing the reason why this should be merged into the main repo.

mahiso:
@rnevet It's only another way to create a docker image instead of using compose. I personally use it to build images automatically on docker.com via the docker/github integration.

You are right, it's not a must have or big reason. Therefore, not problem to close the PR.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:2.7-slim
+MAINTAINER Maik Hinrichs <maik@mahiso.de>
```
rnevet: Your name and email?

rnevet:
@mahiso I'm not against, just asking questions as I need to understand better the motivation.

Dockerfile
mahiso: Right. Why do you ask? Is there a project email address?

rnevet: No, can you set it to GitHub url? or does it have to be an email?

mahiso: I replaced it by a LABEL.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r ./requirements.txt
+
+COPY . .
+
+VOLUME /data
+
+RUN ln -s /data/market_data market_data; \
+    ln -s /data/log/botlog.json www/botlog.json
```
rnevet: does this require specific configuration?

mahiso: No, no configuration changes in config, web server or log json. All this are located in /data (could be a extra docker data volume or a local dir) and are safe from container changes (e.g remove and create a new one using new image version)
Directory structure in /data is:
/data
    |- marked_data
    |       | - marked data files
    |- log
    |       | - botlog.json
    |- config
    |        | - default.cfg

rnevet:
@laxdog @Evanito any additional input?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Build: docker build -t <your_id>/pololendingbot .
+# Run: docker run -d -v /pololendingbot_data:/data -p 8000:8000 <your_id>/pololendingbot
+#
+
+WORKDIR /usr/src/app
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r ./requirements.txt
+
+COPY . .
```
laxdog: Why not just copy everything to the workdir instead of messing about with symlinks and config paths.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+#
+# Build: docker build -t <your_id>/pololendingbot .
+# Run: docker run -d -v /pololendingbot_data:/data -p 8000:8000 <your_id>/pololendingbot
+#
+
+WORKDIR /usr/src/app
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r ./requirements.txt
```
laxdog: Should this be CMD so the requirements get updated? Maybe not, I'm not sure. I do think all the RUNs should be on one line though to reduce the number of layers in the image.

mahiso: Reason is that I dislike change dependencies after image creation during each start of the container. Never touch a running system...

Dockerfile
mahiso: Holding data in an extra volume protects it all for container changes. Never hold data in application container!

Dockerfile
laxdog: OK, fair enough.

Dockerfile
laxdog: OK. I'll merge this then.

PR #432, opened 2017-07-28
prevents directory listing and redirects to lendingbot.html

mahiso:
Prevents directory listing and redirects to lendingbot.html.

## Types of changes

## Checklist:


tamaskan:
are you sure that not having a index.html-file is not intentional ?

rnevet:
@tamaskan Sadly I don't remember why I decided not to add an index.html or to actually name the file index.html. Might have had an idea but can't think of one now.

langerhans:
I'd suggest just renaming the lendingbot.html file to index.html. Given the very specific filename it is otherwise really easy to find running instances of the bot with a simple Google search. I also recommend adding a robots.txt to prevent crawlers from indexing the bot site regardless. The contents should be 
```
User-agent: *
Disallow: /
```
Otherwise there are still easy ways to find the bot.

tamaskan:
there is only 1 reason i can imagine: you put the lendingbot.html beside your important index.html and upon a forced git pull the important index.html get replaced

rnevet:
@tamaskan well this PR will create an index.html - which means that won't be possible anymore.
Ah! I remember, reason was I was running it locally (without a webserver), so I gave it a meaningful name.

tamaskan:
mystery solved :-)

PR #498, opened 2017-08-29
Web UI support for plugins

utdrmac:
Allows plugins to have their own HTML page within the web UI by calling `self.log.addSectionlog('plugins', '<pluginName>', 'webui', True);`

This will add a navbar element for the plugin using <pluginName>.html as the link.

## Types of changes


utdrmac:
Depends on #497

www/lendingbot.js
```diff
@@ -39,6 +39,41 @@ function updateJson(data) {
...
+    $.each($('#navbar-menu li'), function() {
+        if ($(this).attr('id') != undefined) {
+            navbarItems.push($(this).attr('id').toLowerCase())
+        }
+    });
+
+    // iterate through enabled plugins; look for 'navbar': true
+    $.each(enabled, function(i, val) {
+        var v = val.toLowerCase()
+        if ((v in plugins) && ('navbar' in plugins[v]) && (plugins[v]['navbar'])) {
```
rnevet: am I missing something, in the documentation you mentioned 'webui' here you are looking for 'navbar' ?

utdrmac: ????? woah. Umm. Lemme investigate. Not sure how that happened. I've got three different git repos for this bot so probably copied wrong.

Issue #489, opened 2017-08-25
Plugins should return true/false on their init


utdrmac:
When a plugin is initialized, it should return true/false to indicate to the pluginManage that it successfully initialized. If false, do not add it to the plugins array.

rnevet:
As we have 1 plugin so far, what's the need?  
Theoretically a plugin can manage this logic internally, no?

utdrmac:
I'm developing another plugin. That'll make 2 :)

I guess, technically, yes, a plugin can have a `self.enabled` and check it every time.

rnevet:
I suggest you make a PR for the improvement of the PluginManager.

PR #506, opened 2017-09-01
Only deal with plugins that successfully initialized

utdrmac:
Plugins can fail to load due to whatever dependencies may be in their code (ex: missing file, dependency on another plugin, etc). If a plugin fails to load, don't add it to the main loop.

## Types of changes


utdrmac:
Fixes #489

Issue #439, opened 2017-08-08
Add prefix on notifications


pedrofaraco:
With the Bitfinex PR, it would be very useful to have a prefix added to notifications, to be able to differentiate between bot instances when they share the same Telegram channel, for example.

![image](https://user-images.githubusercontent.com/29162354/29078262-364127da-7c30-11e7-8e6b-39378f4302d5.png)

The above image shows two Telegram notifications. The first one is from Poloniex. The second, from Bitfinex.

A prefix setting in config file would allow something like this:
`[Polo] Lent: [0.8558 BTC @ 3.9665%]`
`[Bitf] Lent: [1.2490 BTC @ 0.0168%]`

Or even like this:
`[Mine] Lent: [0.8558 BTC @ 3.9665%]`
`[Kids] Lent: [1.2490 BTC @ 0.0168%]`

Evanito:
@laxdog 
Could you see if this can be implemented into Notifications?

rnevet:
The string is actually generated in the Logger.

mahiso:
Why not in:
https://github.com/BitBotFactory/poloniexlendingbot/blob/e0f42abbe50d5f5d7dbdca2d70c5af537aaeb4bf/modules/Notify.py#L116

A new config vaiable
notifiy_prefix = [what ever]

and *pseudo code*:
msg = notify_prefix ? notify_prefix + msg : msg

utdrmac:
@pedrofaraco Take a look at my PR and give it a test run if you can.

pedrofaraco:
Works like a charm!

![image](https://user-images.githubusercontent.com/29162354/29999320-47092ba0-901a-11e7-830a-eef5fe46d122.png)

Many thanks!

utdrmac:
@pedrofaraco Can you close this issue? Seems the merge didn't take care of that part. Thanks!

PR #507, opened 2017-09-02
Add notification prefix

utdrmac:
Allows the use of a prefix string for all notifications

## Types of changes

screen shot 2017-09-01 at 9 48 57 pm


utdrmac:
Fixes #439

PR #543, opened 2017-09-27
Log bot "version" on error handling

rnevet:
When logging an exception also log the number of commits on current branch, will allow to easily identify if reporter is on master and updated.

## TESTING STAGE
in progress

## Types of changes

## Checklist:

- [n/a] **I have updated the documentation in /docs if I have changed the config, arguments, logic in how the bot works, or anything that understandably needs a documentation change.**
- [n/a] **I have updated the config file accordingly if my change requires a new configuration setting or changes an existing one.**


modules/Configuration.py
```diff
@@ -241,3 +241,15 @@ def get_plugins_config():
     if config.has_option("BOT", "plugins"):
         active_plugins = map(str.strip, config.get("BOT", "plugins").split(','))
     return active_plugins
+
+
+# gets the git commit count as version for master.
+def get_bot_version():
```
Evanito: This function belongs in the Data module, it has nothing to do with the Config.

laxdog:
Good idea, we definitely need this. 

Was just thinking there, we could add a second place to look up the version number if it's not a git repo. Then we can at least manually update that every once in a while, it could be the actual release version.

rnevet:
@Evanito If the git command doesn't return a number a default version number '3.0.0' will be returned. We can update this number from time to time, so we have some indication of the bot version. 

I currently don't have time to include the CLI command, can I merge without?

modules/Data.py
```diff
@@ -141,3 +141,14 @@ def truncate(f, n):
...
+
+
+# gets the git commit count as version for master.
+def get_bot_version():
+    import subprocess
+    output = subprocess.check_output(["git", "rev-list","--count", "HEAD"])
+    try:
+        int(output)
+        return output
+    except ValueError:
```
laxdog: You need to catch 'Exception' here I think. Or work out the other exceptions. 

If it's not a git repo, output won't exist and it'll be a NameError, if git doesn't exist it'll be an OSError. Not sure what else there could be.

rnevet: Fixed.

PR #554, opened 2017-10-04
Use FRR on bitfinex as min_daily_rate

laxdog:
Use the flash return rate rather than the min daily rate set for the coin
## Description

Seems to work quite well from testing. You can manage to get either side of the FRR wall using spreads and it helps MACD work a little better too if you don't have time to constantly fix the min_daily_rate for each currency.

Need to add some config around it to allow it on and off for each coin, but raising now in case anyone wants to comment on it or test it. 

This PR is so messy as it's branched from the un-merged ERR_RATE_LIMIT_fix branch. So bear that in mind if testing. 

It will assume that you want to use the FRR for your min on every coin that you're lending.

I might look at making it use the FRR if it's higher than a value you already have set for the min_daily, I'm guessing that's what most people will want. 

## Types of changes

## Checklist:

### DIFF
--- a/modules/Lending.py
+++ b/modules/Lending.py
@@ -58,7 +58,7 @@ def init(cfg, api1, log1, data, maxtolend, dry_run1, analysis, notify_conf1):
     global sleep_time, sleep_time_active, sleep_time_inactive, min_daily_rate, max_daily_rate, spread_lend, \
         gap_bottom_default, gap_top_default, xday_threshold, xday_spread, xdays, min_loan_size, end_date, coin_cfg, \
         min_loan_sizes, dry_run, transferable_currencies, keep_stuck_orders, hide_coins, scheduler, gap_mode_default, \
-        exchange, analysis_method, currencies_to_analyse 
+        exchange, analysis_method, currencies_to_analyse, all_currencies
 
     exchange = Config.get_exchange()
 
@@ -81,6 +81,7 @@ def init(cfg, api1, log1, data, maxtolend, dry_run1, analysis, notify_conf1):
     min_loan_sizes = Config.get_min_loan_sizes()
     dry_run = dry_run1
     transferable_currencies = Config.get_currencies_list('transferableCurrencies')
+    all_currencies = Config.get_all_currencies()
     currencies_to_analyse = Config.get_currencies_list('analyseCurrencies', 'MarketAnalysis')
     keep_stuck_orders = Config.getboolean('BOT', "keepstuckorders", True)
     hide_coins = Config.getboolean('BOT', 'hideCoins', True)
@@ -242,7 +243,8 @@ def lend_all():
             break
     try:
         for cur in lending_balances:
-            usable_currencies += lend_cur(cur, total_lent, lending_balances, ticker)
+            if cur in all_currencies:
+                usable_currencies += lend_cur(cur, total_lent, lending_balances, ticker)
     except StopIteration:  # Restart lending if we stop to raise the request limit.
         lend_all()
     set_sleep_time(usable_currencies)
--- a/modules/Lending.py
+++ b/modules/Lending.py
@@ -64,18 +64,17 @@ def init(cfg, api1, log1, data, maxtolend, dry_run1, analysis, notify_conf1):
 
     sleep_time_active = float(Config.get("BOT", "sleeptimeactive", None, 1, 3600))
     sleep_time_inactive = float(Config.get("BOT", "sleeptimeinactive", None, 1, 3600))
-    min_daily_rate = Decimal(Config.get("BOT", "mindailyrate", None, 0.003, 5)) / 100
-    max_daily_rate = Decimal(Config.get("BOT", "maxdailyrate", None, 0.003, 5)) / 100
+    exchangeMax = 7 if exchange == 'BITFINEX' else 5
+    min_daily_rate = Decimal(Config.get("BOT", "mindailyrate", None, 0.003, exchangeMax)) / 100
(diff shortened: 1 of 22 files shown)


laxdog:
Just thought, it would probably be useful to have a delta with it too.

utdrmac:
Cool on this. I've noticed many of my loans in bitfinex are being calculated well below FRR, which to me means I'm possibly missing out on some % interest.

laxdog:
Yip, I noticed exactly the same, that's why I started looking into it. 

Seems to make quite a difference actually. I haven't had many coins sitting unused for any length of time. 

I'll try and get the rest of it finished shortly and merge. Need to merge that other branch first though.

laxdog:
Thanks for the feedback and the code @MoonMaker 

If you want to submit your own PR then you will need to fork this repository to your own github page. Make the changes there, then submit a pull request (selecting forks).

You might want to wait until I'm finished with this though, as there have been a few other changes which impact this PR.

utdrmac:
Could you add some output/logging that states something like "Used FRR due to calculated rate being < minloanrate"?

laxdog:
Yea, I'll add more logging @utdrmac need to merge the delta stuff first.

utdrmac:
While you're at it, could you switch logging in some way to the web UI? I found [this question](https://www.reddit.com/r/poloniexlendingbot/comments/76uc8t/new_install_how_do_i_tell_if_marketanalysis/?st=j90ixewz&sh=90fe8c9c)

rnevet:
I suspect that if a coin doesn't have a coin cfg it will not use frrasmin after latest changes.

modules/Lending.py
```diff
@@ -256,10 +256,15 @@ def get_min_daily_rate(cur):
                 max_active_alerted[cur] = True
                 log.log('maxactive amount for ' + cur + ' set to 0, won\'t lend.')
             return False
-        cur_min_daily_rate = Decimal(coin_cfg[cur]['minrate'])
+        if exchange == 'BITFINEX' and coin_cfg[cur]['frrasmin']:
+            frr_rate = Decimal(api.get_frr(cur)) + ( Decimal(coin_cfg[cur]['frrdelta']) / 100 )
```
utdrmac: Why is frrdelta added to the current frr? Wouldn't that make your loan LESS likely to be picked since your rate would be higher than FRR? Additionally, why is the frrdelta being divided by 100? If I set frrdelta to 0.0001, then it's divided by 100 to = 0.000001 which is added to the FRR of (at this time 0.0019), that gives my frr_rate = 0.001901

But don't I want 0.001899 so that I am lower than FRR?

utdrmac:
I think @rnevet has a point. I have neither `coinconfig` or per-coin `[  ]` settings. I just have globals to be used for all coins.

utdrmac:
Actually, this entire if-block is being skipped because I have neither per-coin nor coinconfig. https://github.com/BitBotFactory/poloniexlendingbot/blob/d1a89be8223dba284ddd8c93bbd2a4fd5f0cb1d2/modules/Lending.py#L253

utdrmac:
What about the case where frr < mindailyrate? Right now, FRR for BCH is 0.00189 and my mindailyrate is 0.004 (cause docs say 0.0031 isn't worth it). Yes, FRR is lower but looking at the spread, 0.004 isn't even on the page. Wouldn't you rather the bot be lending than not?

Should I just set mindailyrate=0 and let FRR+frr_delta always be my minimum?

utdrmac:
`WARN: [BOT]-mindailyrate's value: '0' is below the minimum limit: 0.003, which will be used instead.` So, can't even use FRR here.

utdrmac:
[This patch](http://termbin.com/z7mp) should fix a few things, mainly using global settings if you don't have per-coin settings. It also adds some logging messages to UI so you know what the bot is choosing. Sorry about the patch, I didn't know how to PR on top of another PR.

rnevet:
@utdrmac lending for such a low interest isn't worth having your funds locked and in the exchange even for that matter.

modules/Lending.py
MoonMaker: Hi, please use the Bitfinex interface to understand FRRdelta. The same approach is giving here. api.get.frr is already coin/100 so you need to do it for frrdelta, too. Later it will multiply with 100.

MoonMaker:
@rnevet: What do you think about to create a more flexible system in order to choose several  algorithms, based on the current funding walls? E.g. in common way, FRR is a good method, when we have a pump then MACD is very good and when we have a dump MAX of last 5min. is working well. 

This could e.g. be solved by defining an interface where the functions are in a folder and can be selected by configuration.

rnevet:
@MoonMaker I'm all for abstracting having the Market Analysis module and having the Lending module only place the offers... I've discussed this with @laxdog  and @Evanito  in the past.

rnevet:
I have to mention that using FRR as min daily rate has seriously reduced my lendings, maybe it should like MA be used as the lowest rate to place a loan and not min daily rate.

rnevet:
@laxdog where is this standing?

laxdog:
I'll pick it back up, been busy of late. I can address most of the issues, but I won't be able to implement the changes between choosing algorithms.

I like that and think we should do, but there's a lot of work in it. There are so many other problems that need fixed before starting into that in my opinion. 

Anyway, I'll see what I can do with this first.

rnevet:
Regarding the "changes between choosing algorithms" of course this isn't in the scope of this change.

laxdog:
Right, I think this is done if someone could have a look at it?

PR #570, opened 2017-10-24
Added Blacklist Option

MoonMaker:
Based on new Bitcoin Gold coin listing on Bitfinex, we receive the following error message:
Error API Error 400: {"message":"Unknown currency"} Requesting https://api.bitfinex.com/v1/lendbook/BTG?limit_asks=100&limit_bids=100
Caught API Error 400: {"message":"Unknown currency"} Requesting https://api.bitfinex.com/v1/lendbook/BTG?limit_asks=100&limit_bids=100 reading from exchange API, ignoring

## Description
The issue is that the coin is listing in Symbols but not in lending book. So, here you can find a solution for this bug. On the other side, I'm already missing a good blacklist feature, because this bot is requesting bitfinex api for coins which are not included in "all_currencies" (in config).

## TESTING STAGE##
Ongoing

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -192,7 +192,7 @@ Blacklist Currencies from Lending
 
 With this option you can create a blacklist of currencies, which you don't want to use at all.
 
-- ``blacklistCurrencies`` is a list of currencies you would like to be avoid.
+- ``blacklistCurrencies`` is a list of currencies you would like to be avoided.
 
     - Default value: Commented out
     - Format: ``BTC,BTS,CLAM,DOGE,DASH,LTC,MAID,XMR,XRP,ETH,FCT``
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -187,6 +187,17 @@ If you regularly transfer funds into your Poloniex account but don't enjoy havin
     - Do not worry about duplicates when using ``ACTIVE``, they are handled.
     - Coins will be transferred every time the bot runs (60 seconds by default) so if you intend to trade or withdrawal it is recommended to turn off the bot or disable this feature.
 
+Blacklist Currencies from Lending
+-----------------------------------
+
+With this option you can create a blacklist of currencies, which you don't want to use at all.
+
+- ``blacklistCurrencies`` is a list of currencies you would like to be transferred.
+
+    - Default value: Commented out
+    - Format: ``BTC,BTS,CLAM,DOGE,DASH,LTC,MAID,XMR,XRP,ETH,FCT``
+    - Commenting it out will disable the feature.
+
 Unimportant settings
 --------------------
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -192,7 +192,7 @@ Blacklist Currencies from Lending
 
 With this option you can create a blacklist of currencies, which you don't want to use at all.
 
-- ``blacklistCurrencies`` is a list of currencies you would like to be transferred.
+- ``blacklistCurrencies`` is a list of currencies you would like to be avoid.
 
     - Default value: Commented out
     - Format: ``BTC,BTS,CLAM,DOGE,DASH,LTC,MAID,XMR,XRP,ETH,FCT``
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -54,7 +54,11 @@ Exchange Sections
 There is a section for each exchange to configure exchange specific attributes.
 
 - ``all_currencies`` List of all supported currencies for funding. The list have to change only
-  when the exchange adds a new supported currency or removes one.
+  when the exchange adds a new supported currency or removes one. You can create a blacklist of currencies by using '#' in front of a specific one.
+
+    - Format: ``BTC,BTS,CLAM,DOGE,DASH,LTC,MAID,XMR,XRP,ETH,FCT,#BTG``
+
+
 
 Timing
 ---------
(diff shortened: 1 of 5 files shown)


mike4001:
Working for me => Thanks !

docs/configuration.rst
```diff
@@ -187,6 +187,17 @@ If you regularly transfer funds into your Poloniex account but don't enjoy havin
     - Do not worry about duplicates when using ``ACTIVE``, they are handled.
     - Coins will be transferred every time the bot runs (60 seconds by default) so if you intend to trade or withdrawal it is recommended to turn off the bot or disable this feature.
 
+Blacklist Currencies from Lending
+-----------------------------------
+
+With this option you can create a blacklist of currencies, which you don't want to use at all.
+
+- ``blacklistCurrencies`` is a list of currencies you would like to be avoid.
```
utdrmac: "would like to avoid" or "would like to be avoided"

utdrmac:
Instead of blacklist, wouldn't the better fix be to make the bot only concern itself with `all_currencies` ?

utdrmac:
As a side note, this PR is working for me too.

MoonMaker:
Yes,  utdrmac. Currently I',m working on to use both and the first test is ongoing.

rnevet:
I think all_currencies should be enough to cover the configuration, in addition you can specify a specific coin you don't want to lend in coin config. The reason you are seeing this error is because you have BTG in your lending account and it isn't lendable - which is "weird" and was only created because of the BTG airdrop.

MoonMaker:
Hi rnevet, normally you have always an option like "all" (here all_currencies) and a blacklist. The management is more easier than starting to optimize all_currencies. E.g. More than 1000 currencies are existing and you want to ignore 10? This is to much work....

rnevet:
@laxdog @Evanito  opinion?

rnevet:
Adding BTG to the all_currencies is also just wrong as this isn't a "supported currency" for lending.

MoonMaker:
Yes, I added it there for testing, because this was exactly the bug. BTG was not in all_currencies but a request is done for BTG lending, because the BTG was in funding wallet. 

In other words: 

- All_currencies had no effect -> Bug closed (this also reduced ERROR_LIMIT messages)
- Blacklist added for easier management
- I will remove BTG from all_currencies but I'm pretty sure it will be added next month on bitfinex (see BCash)

mahiso:
#571: whitelisting.

modules/Lending.py
```diff
@@ -81,6 +81,8 @@ def init(cfg, api1, log1, data, maxtolend, dry_run1, analysis, notify_conf1):
     min_loan_sizes = Config.get_min_loan_sizes()
     dry_run = dry_run1
     transferable_currencies = Config.get_currencies_list('transferableCurrencies')
+    all_currencies = Config.get_list(exchange, 'all_currencies')
```
Evanito: Use `Config.get_all_currencies()`, which has better situation handling and was specifically made for returning this value.

modules/Lending.py
```diff
@@ -242,7 +244,8 @@ def lend_all():
             break
     try:
         for cur in lending_balances:
-            usable_currencies += lend_cur(cur, total_lent, lending_balances, ticker)
+            if cur in all_currencies and cur not in blacklist_currencies:
```
Evanito: As I will say in my final review, a whitelist system with `all_currencies` is sufficient for determining if a coin is to be lent.

modules/Bitfinex.py
```diff
@@ -30,6 +30,9 @@ def __init__(self, cfg, log):
         self.symbols = []
         self.ticker = {}
         self.tickerTime = 0
+	self.baseCurrencies = ['USD', 'BTC', 'ETH']
```
Evanito: I like what you did in the Bitfinex module, making it easy to add new base currencies is a useful change.

modules/Lending.py
Evanito: Since I cannot comment directly on `get_all_currencies()`:

I would check for a sentinel value before any currencies, and if it is present, ignore the currency.
For example, with a sentinel character of `#`:
```
def get_all_currencies():
    '''
    Get list of all supported currencies by exchange
    '''
    exchange = get_exchange()
    if config.has_option(exchange, 'all_currencies'):
        cur_list = []
        raw_cur_list = config.get(exchange, 'all_currencies').split(',')
        for raw_cur in raw_cur_list:
            cur = raw_cur.strip(' ').upper()
            if (cur[0] != "#"):
                cur_list.append(cur)
        return cur_list
    elif exchange == 'POLONIEX':
        # default, compatibility to old 'Poloniex only' config
        return ['STR', 'BTC', 'BTS', 'CLAM', 'DOGE', 'DASH', 'LTC', 'MAID', 'XMR', 'XRP', 'ETH', 'FCT']
    else:
        raise Exception('ERROR: List of supported currencies must defined in [' + exchange + '] all_currencies.')
```

This way, we can keep the whitelist while also allowing temporary blacklisting.

MoonMaker: Ok lets use #. Its not perfect because of:
all_currencies = BTC,BTG,#BTG <- BTG will be used 

but it allows more flexibility then nothing.

modules/Lending.py
MoonMaker: Yep, I had problems with all_currencies. Maybe, I did a mistake at this point in time. So, I will use it again to look if it is working, now.

docs/configuration.rst
```diff
@@ -54,7 +54,11 @@ Exchange Sections
 There is a section for each exchange to configure exchange specific attributes.
 
 - ``all_currencies`` List of all supported currencies for funding. The list have to change only
-  when the exchange adds a new supported currency or removes one.
+  when the exchange adds a new supported currency or removes one. You can create a blacklist of currencies by using '#' in front of a specific one.
```
Evanito: Better grammar:

You can blacklist specific currencies by prefacing it with a #, this is the same as not including it on the list.

PR #571, opened 2017-10-25
Currency Whitelisting

mahiso:
NOT READY

Currency Whitelisting:

Bot will manage only currencies listed in "all_currencies" configuration attribute.

ToDos:
- Poloniex implementation
- Documentation

## Types of changes

## Checklist:

### DIFF
--- a/modules/Bitfinex.py
+++ b/modules/Bitfinex.py
@@ -137,14 +137,17 @@ class Bitfinex(ExchangeApi):
         https://bitfinex.readme.io/v1/reference#rest-auth-offers
         """
         bfx_resp = self._post('offers')
-        resp = Bitfinex2Poloniex.convertOpenLoanOffers(bfx_resp)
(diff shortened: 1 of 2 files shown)


rnevet:
Isn't this a duplicate of #570?

mahiso:
@rnevet: No, its not. #570 implemented a blacklisting, this PR is a whitelisting, so that the bot manage only currencies defined in all_currencies. So it is possible to lend by bot only a subset of currencies and others by hand. At now PR implements it for Bitfinex only, Poloniex will follow.

rnevet:
@mahiso the whitelisting part is also included in the #570 implementation and as far as I can see for both platforms, no?

rnevet:
Also I don't think this should be introduced on the API implementation level.

mahiso:
The implementation at API level has the approach that it is not needed to care about it before every call. The API decides what currency (in all of calls and returns(!)) is allowed and which not.

rnevet:
@mahiso #570 enforces whitelist for lending, which should be enough. Please review.

PR #588, opened 2017-12-05
Allow loan history DB name to be configured

laxdog:
## Description

Allows multiple bots to run from the same directory, for example in a docker setup. 

## TESTING STAGE

Testing

## Types of changes

## Checklist:

### DIFF
--- a/default.cfg.example
+++ b/default.cfg.example
@@ -130,6 +130,7 @@ hideCoins = True
 
 #[ACCOUNTSTATS]
 #ReportInterval = 86400
+#database_name = loan_history
 
 #[CHARTS]
 #DumpInterval = 21600
--- a/plugins/AccountStats.py
+++ b/plugins/AccountStats.py
@@ -33,6 +33,7 @@ class AccountStats(Plugin):
         self.init_db()
         self.check_upgrade()
         self.report_interval = int(self.config.get("ACCOUNTSTATS", "ReportInterval", 86400))
+        self.database_name = self.config.get("ACCOUNTSTATS", "database_name", "loan_history")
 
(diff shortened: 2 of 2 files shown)


rnevet:
This doesn't cover all files used by the bot so it won't achieve the goal. For example the plugins files for AccountStats.

laxdog:
I don't understand. Isn't the only file that's written to the loans_history one @rnevet ?

I can't see any other files in the AccountStats.py

rnevet:
It is also used for charts plugin.
and what about the file for MarketAnalysis?

I'm not sure this approach is maintainable, maybe create a contained output folder?

laxdog:
The market analysis data is actually best kept as the same file, that means every instance gets access to the data the others are writing. This is how I'm running it currently. That's why I added the exchange prefix to those files.

I can add the market_data folder as well if you want, but this particular file needs to be configurable too.

I see the reference in Charts.py, I'll make that change.

rnevet:
I suggest we have shared_data , out instance_data folders... which will allow both the required behaviors..
How are you going to sync access the shared_data?

laxdog:
I feel this is getting overly complicated for a simple change to allow a filename to be configurable. I'll try to get around to it. I'm still in the middle of the FlaskServer and FRR as min branches though.

The access is already synced if you're using the docker-compose file. It mounts the local dir as a volume and then links it into the working dir of the container so you have access to the market data folder.

rnevet:
I agree, I thought it might be complicated.

laxdog:
It's going to have to wait until after the flask stuff is finished. There will be too many conflicts that I don't want to have to resolve.

PR #602, opened 2018-01-02
[UPDATE] Adding code to store logs

vinnyfs89:
## Types of changes

## Checklist:

- [ ] **I have read CONTRIBUTING.md**
- [ ] **I fully understand [Github Flow.](https://guides.github.com/introduction/flow/)**
- [ ] **My code adheres to the [code style of this project.](https://poloniexlendingbot.readthedocs.io/en/latest/contributing.html)**

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -64,6 +64,24 @@ To download the bot you can either:
 
         [Install]
         WantedBy=multi-user.target
+        
+    To store logs :
+    
+    .. code-block:: text
+
+        [Unit]
+        Description=LendingBot service
+        After=network.target
+
+        [Service]
+        Type=simple
+        ExecStart=/usr/bin/python <INSTALLATION DIRECTORY>/lendingbot.py  > <INSTALLATION DIRECTORY>/lendingbot.log 2>&1
+        WorkingDirectory=<INSTALLATION DIRECTORY>
+        RestartSec=10
+        Restart=on-failure
+
+        [Install]
+        WantedBy=multi-user.target
 
     Credit to GitHub user utdrmac.
 


utdrmac:
Shouldn't the logs be printed to stdout/stderr so that they are captured, and logged, and rotated by journalctl?

vinnyfs89:
I just add another way to do this.

utdrmac:
But then doesn't this break established functionality provided by the OS to do exactly what you want? Seems like re-inventing the wheel. You already have a mechanism for storing, accessing and rotating the logs, journalctl. or am I missing something more?

vinnyfs89:
In fact, I disagree with your point of view. For my scenario, although you do not agree, as the python script is initialized as a service in background mode, no information about the script will be displayed, so for my scenario storing the logs becomes something useful. I do not see how "Reinventing the wheel" ...

But it's just my point of view that will probably help other people, but you can deny the pull request if you want.

rnevet:
@utdrmac can you maybe clarify how should you use journalctl to view the bots log in the service scenario? (I have to admit Unix isn't my domain) maybe add that to the docs?

utdrmac:
Since systemd became process 1 in all major linux operating systems, things like auto-restarting crashed applications, handling logs, rotating logs, etc, all of that is built-in to systemd. You simply launch your application and systemd does the rest. You don't need to code in external logging logic. Just print everything to stdout and systemd captures it all automatically and sends it to journald. You can view the logs like `journalctl -u lendingbot`

utdrmac:
vinnfs's patch is essentially redirecting stdout and stderr away from systemd capture, and sending to a separate file. It's a simple solution to have ONLY the lendingbot logs in a separate file. You can get the same exact info by letting systemd handle it and use the journactl command to view the logs, albeit learning how to properly use journalctl.

rnevet:
Thanks @utdrmac!

Issue #314, opened 2017-05-02
Feature Request: Version checking & notification


utdrmac:
Lots of changes are being made to this bot (Yay!) but subscribing to every issue is cluttering up my Inbox (Boo!). I'm not sure if I can subscribe to only "releases" or only to be notified when commits/merges on/to the master branch happen.

To that effect, I would like to see a daily/weekly version checker that simply prints a NOTE in the web gui and optionally, sends a notification using the new notifications additions that a new version is available.

Evanito:
Theoretically possible using a version file and a check between the local and remote versions.

rnevet:
I would like to have it automated, related to master commits.

laxdog:
Should probably be done through Travis if you want it automated.

utdrmac:
Version checking should be related to master commits. A thread/timer should check once a day/week/<configurable> that the current checkout is different from the current master.

rnevet:
We could check with a git command in the python script, however this will only work if the bot is running from a git folder and not a plain copy.

utdrmac:
Ah yes, if someone downloads the .zip/.tgz, there won't be any git-related stuff. So I guess there would need to be a version.txt file with the current commit hash that checks against the master branch copy at github.com?

laxdog:
I've started some work on this. I'm adding a packaging branch where I'm hoping to address versioning, PyPi support and Travis builds with testing. Still a long way off, but it's being looked at.

Evanito:
We could move to a tag system and keep everything on the github side.
Compare the local tags to the remote tags, and notify if a discrepancy is found.

Getting remote tags: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-ls-remote.html

PR #623, opened 2018-04-06
Autoupdate

Evanito:
## Description

- When ran using the `-up` argument, will perform a git pull before starting the bot normally.
- Tried to be user friendly with error messages.
- Didn't want to reinvent the wheel, so let git pull decide when an update is needed.
- Requires you to restart the bot to update (using cron or manually)
- [ ] Need to add new arg to docs

Thoughts?:
- [ ] Can probably be changed to autorestart self once per time frame?
- [ ] Allow usage through config instead of arg?? (Config is for accounts imo, args are for the software)

#314 

## Types of changes

## Checklist:

### DIFF
--- a/lendingbot.py
+++ b/lendingbot.py
@@ -40,6 +40,7 @@ if update:
     try:
         up_out = subprocess.check_output(up_cmd).decode(sys.stdout.encoding)
     except:
+        print "Looks like you need to use sudo..."
         up_cmd.insert(0, "sudo")
         up_cmd.insert(1, "--non-interactive")
         up_out = subprocess.check_output(up_cmd).decode(sys.stdout.encoding)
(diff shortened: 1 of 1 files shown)


rnevet:
I'm not sure if this should be solve as part of the bot code. For example for my setup it wouldn't work as I'm redirecting the outputs and the new cmd will not take those output settings I think.

Evanito:
I also already have autoupdating built into my cron job...
This is honestly just a solution for those who don't have advanced setups, still allows users to update another way.

lendingbot.py
```diff
@@ -27,10 +27,51 @@
...
+        print "You do not have git installed!"
+    elif "a git repository" in up_out:
+        print "You did not install the bot using git!\n" \
+              "Reinstall using git using:\n" \
+              "(sudo) git clone http://github.com/BitBotFactory/poloniexlendingbot.git"
+    elif "Permission denied" in up_out:
+        print "Your sudo is not configured for non-interactive usage, this is necessary."
+    else:
+        print up_out
+    cmd = ["python", "lendingbot.py"]
```
rnevet: Maybe you could use `argv` instead of rebuilding the request?

